PR #75, opened 2015-01-26
Catch event stop during create/update

amenophis:
related to https://github.com/Sylius/Sylius/issues/2022


amenophis:
@pjedrzejewski @stloyd what do you think about this ?
Better solution ?

stloyd:
This cant be merged here as it's read-only split. But in overall looks good.

PR #79, opened 2016-01-22
Fallback on metadata default form class 

pdias:
In route, if don't exist the parameter "form", getFormType returns a string name for the form. A deprecated warning occurs (Accessing type "*****" by its string name is deprecated since version 2.8 and will be removed in 3.0. Use the fully-qualified type class name "************").


jrobeson:
this repository does not accept pull requests, you have to fill them in main sylius repo

PR #80, opened 2016-07-11
Fixed windows path bug

caponica:
Updated driver detection to work with Windows file paths, which have "\" instead of "/".


jrobeson:
pull requests aren't accepted here, add them to the main sylius project.

PR #92, opened 2019-06-11
Removing '2' from 'Symfony2'

loevgaard:
Seems outdated with the '2' after 'Symfony' and that is certainly not the case :)


pamil:
Thanks, Joachim! :1st_place_medal:

loevgaard:
The small details count ;)

PR #97, opened 2019-06-13
Autodiscover resource's model interfaces and deprecate explicit configuration

### LINKED ISSUE #94: Doctrine class metadata not found for extended interface
**Sylius version affected**: 1.x

**Description**  
For any resource, it should be possible to get the Doctrine class metadata using any of the interfaces for the resource defined at the component / core / app level.

**Steps to reproduce**  
1. Create a new project from Sylius-Standard.
2. Try to get Doctrine class metadata using `Sylius\Component\Core\Model\ProductInterface`.

**Possible Solution**  
I think there is a design problem.

If you look at https://github.com/Sylius/SyliusResourceBundle/blob/v1.5.0/src/Bundle/DependencyInjection/Compiler/DoctrineTargetEntitiesResolverPass.php

it does not go up the inheritance chain - and it's not safe to do so, as we can't tell which interfaces should be mapped, and which should not. So if we set `interface` to a child interface, things would stop working if you try using the parent interface(s).

What is actually expected is that all of the interfaces should work, e.g.:

- `Sylius\Component\Product\Model\ProductInterface`
- `Sylius\Component\Core\Model\ProductInterface`
- `App\Model\ProductInterface` (for example)

So we need a way to map multiple interfaces.

pamil:
Fixes #94, solution described in https://github.com/Sylius/SyliusResourceBundle/issues/94#issuecomment-501568904.

### DIFF
--- a/src/Bundle/spec/DependencyInjection/Compiler/Helper/TargetEntitiesResolverSpec.php
+++ b/src/Bundle/spec/DependencyInjection/Compiler/Helper/TargetEntitiesResolverSpec.php
@@ -1,0 +1,103 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace spec\Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler\Helper;
+
+use PhpSpec\ObjectBehavior;
+use Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler\Helper\TargetEntitiesResolverInterface;
+use Sylius\Bundle\ResourceBundle\Tests\Fixtures\AnimalInterface;
+use Sylius\Bundle\ResourceBundle\Tests\Fixtures\Bear;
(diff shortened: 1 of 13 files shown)


src/Bundle/spec/DependencyInjection/Compiler/Helper/TargetEntitiesResolverSpec.php
```diff
@@ -0,0 +1,103 @@
...
+use Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler\Helper\TargetEntitiesResolverInterface;
+use Sylius\Bundle\ResourceBundle\Tests\Fixtures\AnimalInterface;
+use Sylius\Bundle\ResourceBundle\Tests\Fixtures\Bear;
+use Sylius\Bundle\ResourceBundle\Tests\Fixtures\BearInterface;
+use Sylius\Bundle\ResourceBundle\Tests\Fixtures\Fly;
+use Sylius\Bundle\ResourceBundle\Tests\Fixtures\FlyInterface;
+use Sylius\Bundle\ResourceBundle\Tests\Fixtures\MammalInterface;
+use Sylius\Bundle\ResourceBundle\Tests\Fixtures\Resource;
+
+class TargetEntitiesResolverSpec extends ObjectBehavior
```
lchrusciel: missing final :)

PR #99, opened 2019-06-26
Upgrade to WinzouStateMachineBundle ^0.4.1

ahilke:
winzou/state-machine-bundle 0.4.1 fixes an incompatibility with Symfony 4. Allowing users to upgrade the state-machine-bundle make using Symfony 4 easier.


mamazu:
:+1: Although this might be a BC break.

pamil:
We'd need to add a compiler pass which provides BC with 0.3 behaviour (`sm.factory` service etc.)

mamazu:
Or probably an alias will just do.

pamil:
An alias in a compiler pass, since we'd modify services coming from outside resource bundle.

PR #100, opened 2019-07-16
Critiera Applier

mamazu:
In the current version of the plugin if you are using it to generate CRUD routes for an api the filter mechanism doesn't support filtering with only a part of the search term.

In this commit I am trying to fix this.

### DIFF
--- a/src/Bundle/Doctrine/ORM/EntityRepository.php
+++ b/src/Bundle/Doctrine/ORM/EntityRepository.php
@@ -15,6 +15,7 @@ namespace Sylius\Bundle\ResourceBundle\Doctrine\ORM;
 
 use Doctrine\ORM\EntityRepository as BaseEntityRepository;
 use Doctrine\ORM\QueryBuilder;
+use function is_string;
 use Pagerfanta\Adapter\ArrayAdapter;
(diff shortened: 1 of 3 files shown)


mamazu:
In this pull request I just noticed that I could change the logic without a test noticing which is usually a bad sign. But I noticed that the tests only test the status code not the content. Which makes tests that should fail like the one `BookApiTest:it_applies_sorting_for_existing_field` pass because the response code is 200 but the body is empty.

lchrusciel:
Filtration is handled with grid bundle, and the empty result is returned due to lack of grid configuration. From what I see, we have already coupled grid and resource bundle (another question is, do we want it that way), so we should probably add a grid definition for books, so it will be possible to filtrate them or remove obsolete tests. 

Anyway, did you encounter any issue related to this functionality?

mamazu:
These are two different things in my opinion. I am talking about the CRUD Api that is generated with the bundle. This has nothing to do with grids.

lchrusciel:
Not sure, if I will find time to solve this issue, but I truly believe, that adding a grid to this route would solve the problem

mamazu:
Yes, that would also be an idea. I was also surprised about the fact that the grid filter work different than the API filters.

PR #102, opened 2019-07-25
bug #10544: create new resource on update if not found

dimaip:
Resolves: https://github.com/Sylius/Sylius/issues/10544

### DIFF
--- a/src/Bundle/Controller/ResourceController.php
+++ b/src/Bundle/Controller/ResourceController.php
@@ -251,7 +251,7 @@ class ResourceController extends Controller
         $configuration = $this->requestConfigurationFactory->create($this->metadata, $request);
 
         $this->isGrantedOr403($configuration, ResourceActions::UPDATE);
-        $resource = $this->findOr404($configuration);
+        $resource = $this->findOrCreate($configuration);
 
         $form = $this->resourceFormFactory->create($configuration, $resource);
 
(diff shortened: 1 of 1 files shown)


src/Bundle/Controller/ResourceController.php
```diff
@@ -251,7 +251,7 @@ public function updateAction(Request $request): Response
         $configuration = $this->requestConfigurationFactory->create($this->metadata, $request);
 
         $this->isGrantedOr403($configuration, ResourceActions::UPDATE);
-        $resource = $this->findOr404($configuration);
+        $resource = $this->findOrCreate($configuration);
```
vvasiloi: It should probably be conditional, since it's only for PUT requests, the PATCH should return 404 if the resource does not exists.
```
        if ($request->isMethod(Request::METHOD_PUT)) {
            $resource = $this->findOrCreate($configuration);            
        } else {
            $resource = $this->findOr404($configuration);
        }
```

vvasiloi:
This has to be added before 

https://github.com/Sylius/SyliusResourceBundle/blob/master/src/Bundle/Controller/ResourceUpdateHandler.php#L41


```
        if (!$manager->contains($resource)) {
            $manager->persist($resource);
        }
```
Otherwise new resources won't be created.

dimaip:
Ah, I was just wondering why it didn't update anything. Will test.

dimaip:
Duh, I just realised I haven't updated this PR. Feel free to take over or I'll try to find the time to finish it later, after the project launch.

PR #104, opened 2019-08-14
[RFC] Add success flashes before post event

Zales0123:
I don't know is it intended behavior or not 🤔 For me, a natural place to throw a post event is **just before redirect**. With this change, we can, for example, operate on flashed basing on create/updated/deleted resource data (I had such a case 💃)

### DIFF
--- a/src/Bundle/Controller/ResourceController.php
+++ b/src/Bundle/Controller/ResourceController.php
@@ -208,13 +208,16 @@ class ResourceController extends Controller
             }
 
             $this->repository->add($newResource);
+
+            if ($configuration->isHtmlRequest()) {
+                $this->flashHelper->addSuccessFlash($configuration, ResourceActions::CREATE, $newResource);
+            }
+
             $postEvent = $this->eventDispatcher->dispatchPostEvent(ResourceActions::CREATE, $configuration, $newResource);
 
             if (!$configuration->isHtmlRequest()) {
                 return $this->viewHandler->handle($configuration, View::create($newResource, Response::HTTP_CREATED));
             }
-
-            $this->flashHelper->addSuccessFlash($configuration, ResourceActions::CREATE, $newResource);
 
             if ($postEvent->hasResponse()) {
                 return $postEvent->getResponse();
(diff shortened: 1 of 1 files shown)


src/Bundle/Controller/ResourceController.php
```diff
@@ -208,14 +208,15 @@ public function createAction(Request $request): Response
             }
 
             $this->repository->add($newResource);
+
+            $this->flashHelper->addSuccessFlash($configuration, ResourceActions::CREATE, $newResource);
```
lchrusciel: This logic should be dispatched only for html requests

src/Bundle/Controller/ResourceController.php
```diff
@@ -289,6 +292,8 @@ public function updateAction(Request $request): Response
                 return $this->redirectHandler->redirectToReferer($configuration);
             }
 
+            $this->flashHelper->addSuccessFlash($configuration, ResourceActions::UPDATE, $resource);
```
lchrusciel:
```suggestion
            if ($configuration->isHtmlRequest()) {
                $this->flashHelper->addSuccessFlash($configuration, ResourceActions::UPDATE, $resource);
            }
```

src/Bundle/Controller/ResourceController.php
```diff
@@ -493,6 +496,8 @@ public function applyStateMachineTransitionAction(Request $request): Response
             return $this->redirectHandler->redirectToReferer($configuration);
         }
 
+        $this->flashHelper->addSuccessFlash($configuration, ResourceActions::UPDATE, $resource);
```
lchrusciel:
```suggestion
        if ($configuration->isHtmlRequest()) {
            $this->flashHelper->addSuccessFlash($configuration, ResourceActions::UPDATE, $resource);
        }        
```

src/Bundle/Controller/ResourceController.php
```diff
@@ -370,14 +373,14 @@ public function deleteAction(Request $request): Response
             return $this->redirectHandler->redirectToReferer($configuration);
         }
 
+        $this->flashHelper->addSuccessFlash($configuration, ResourceActions::DELETE, $resource);
```
lchrusciel:
```suggestion
        if ($configuration->isHtmlRequest()) {
            $this->flashHelper->addSuccessFlash($configuration, ResourceActions::DELETE, $resource);
        }        
```

src/Bundle/Controller/ResourceController.php
Zales0123: I don't know why I've done this only for the `createAction` XD 🦀

pamil:
Thanks, Mateusz! :tada:

PR #108, opened 2019-09-12
Update willdurand/hateoas-bundle dependency

pierre-H:


pamil:
It looks like we need to support JMS Serializer Bundle ^3.0 as well :/

pamil:
Thanks for the PR, a similar change has been merged in #119.

PR #114, opened 2019-11-28
Updating composer dependencies

mamazu:
The `lakion/api-test-case` package is now replaced with it's successor.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,7 @@
         "akeneo/phpspec-skip-example-extension": "^4.0",
         "doctrine/orm": "^2.5",
         "gedmo/doctrine-extensions": "^2.4",
-        "lchrusciel/api-test-case": "^3.1.0",
+        "lchrusciel/api-test-case": "^4.1.0",
         "matthiasnoback/symfony-dependency-injection-test": "^3.0",
         "phpspec/phpspec": "^5.0",
         "phpstan/phpstan-phpunit": "^0.11",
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -45,7 +45,7 @@
         "akeneo/phpspec-skip-example-extension": "^4.0",
         "doctrine/orm": "^2.5",
         "gedmo/doctrine-extensions": "^2.4",
-        "lakion/api-test-case": "^3.1.0",
+        "lchrusciel/api-test-case": "^3.1.0",
```
lchrusciel: Can we update it to '^4.1' as well?

mamazu: Sure, currently this fails due to some unknown error.

mamazu:
I think the build job failed for some reason. Could you restart it?

lchrusciel:
done

lchrusciel:
Ah, this container was removed from dependencies and added as a suggestion. That's why it's failing.

Ref. https://github.com/lchrusciel/ApiTestCase/blob/master/UPGRADE.md#upgrading-from-v3-to-v4

src/Bundle/spec/Controller/RequestConfigurationFactorySpec.php
```diff
@@ -43,7 +43,7 @@ function it_creates_configuration_from_resource_metadata_and_request(
         $request->headers = $headersBag;
         $request->attributes = $attributesBag;
 
-        $headersBag->get('Accept', null, false)->willReturn([]);
+        $headersBag->get('Accept', null)->illReturn([]);
```
pamil:
```suggestion
        $headersBag->get('Accept', null)->willReturn([]);
```

mamazu:
@pamil This is the new way of doing things. However to preserve backwards compatibilty the `all` function that psalm complains about already has documentation like that.

pamil:
Thanks, @mamazu! :tada:

Issue #121, opened 2019-12-11
Symfony 5 support


pamil:
The issue to track and coordinate efforts.

Roshyo:
Where are we on this ?

loic425:
Can we help on something?

pamil:
There's no work done yet, I'd start from adding a line to `.travis.yml` with `SYMFONY_VERSION=5.0.*`, allowing for Symfony ^5.0 components in `composer.json` and running the tests.

loic425:
@pamil I want to help finishing that big work, but some PR are still on review on SyliusResourceBundle.

alexsoftdev:
Any news about sf5?

loic425:
> Any news about sf5?

it supports Symfony 5 since 1.7.x

PR #135, opened 2020-01-02
Allow for DoctrineBundle ^2.0

pamil:
Related to #121.
The only class we use in ResourceBundle has not been changed in 2.0, so we can support both at the same time.


Zales0123:
Thanks, Kamil! :1st_place_medal:

PR #136, opened 2020-01-02
Remove unneccessary dependency on winzou/state-machine in the component

pamil:


Zales0123:
Thanks, Kamil! :1st_place_medal:

PR #139, opened 2020-01-03
Fix event dispatcher deprecation

dannyvw:
User Deprecated: Calling the "Symfony\Component\EventDispatcher\EventDispatcherInterface::dispatch()" method with the event name as first argument is deprecated since Symfony 4.3, pass it second and provide the event object first instead.


GSadee:
Thank you, Danny! :1st_place_medal:

PR #140, opened 2020-01-03
Switch to FOSRestBundle dev-master for compatibility with Symfony 5

loic425:
@pamil Just a try...


loic425:
Parameters class is not final, but there is no service. Its instance creation uses directly this class.

Sylius\Bundle\ResourceBundle\Controller\RequestConfigurationFactory line 59
`return new $this->configurationClass($metadata, $request, new Parameters($parameters));`

Did you abort this PR due to composer validation error?

PR #149, opened 2020-01-31
Fix DoctrineORMAdapter with $fetchJoinCollection false making result …

### LINKED ISSUE #148: $fetchJoinCollection should be either true by default or let developer decide what should be passed in Sylius\Bundle\ResourceBundle\Doctrine\ORM\EntityRepository::getPaginator
**Sylius version affected**: 1.6.0 at least

**Description**  

The setting shipped in 1.6 can return much less result than expected.

**Steps to reproduce**  
Consider I was trying to get a Pagerfanta for product with specific variant code pattern, with `$fetchJoinCollection` set as false. In `doctrine/orm/lib/Doctrine/ORM/Tools/Pagination/Paginator.php` `getIterator` it will execute something like

```sql
SELECT 
...
FROM product p
JOIN variant pv ON ...
WHERE
...
LIMIT 100
```

Since 1 product can have more than 1 variants, number of distinct products returned is likely be less than 100.

**Possible Solution**  
With `$fetchJoinCollection` set to true, doctrine will first select all distinct IDs first and then select all tuples afterwards to ensure number of products returned matches what `setMaxResult` set.

Long long time ago $fetchJoinCollection was set to true and everything works fine.

tom10271:
…returned less

| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #148 
| License         | MIT

**Description**  

The setting shipped in 1.6 can return much less result than expected.

**Steps to reproduce**  
Consider I was trying to get a Pagerfanta for product with specific variant code pattern, with `$fetchJoinCollection` set as false. In `doctrine/orm/lib/Doctrine/ORM/Tools/Pagination/Paginator.php` `getIterator` it will execute something like

```sql
SELECT 
...
FROM product p
JOIN variant pv ON ...
WHERE
...
LIMIT 100
```

Since 1 product can have multiple variants, number of distinct products returned is likely less than 100.

**Possible Solution**  
With `$fetchJoinCollection` set to true, doctrine will first select all distinct IDs first and then select all tuples afterwards to ensure number of products returned matches what `setMaxResults` set.

Long long time ago $fetchJoinCollection was set to true and everything works fine.

### DIFF
--- a/src/Bundle/Doctrine/ORM/EntityRepository.php
+++ b/src/Bundle/Doctrine/ORM/EntityRepository.php
@@ -57,10 +57,10 @@ class EntityRepository extends BaseEntityRepository implements RepositoryInterfa
         return $this->getPaginator($queryBuilder);
     }
 
-    protected function getPaginator(QueryBuilder $queryBuilder): Pagerfanta
+    // Use output walkers option in DoctrineORMAdapter should be false as it affects performance greatly (see #3775)
+    protected function getPaginator(QueryBuilder $queryBuilder, $fetchJoinCollection = true, $useOutputWalker = false): Pagerfanta
     {
-        // Use output walkers option in DoctrineORMAdapter should be false as it affects performance greatly (see #3775)
-        return new Pagerfanta(new DoctrineORMAdapter($queryBuilder, false, false));
+        return new Pagerfanta(new DoctrineORMAdapter($queryBuilder, $fetchJoinCollection, $useOutputWalker));
     }
 
     /**


src/Bundle/Doctrine/ORM/EntityRepository.php
```diff
@@ -57,10 +57,10 @@ public function createPaginator(array $criteria = [], array $sorting = []): iter
         return $this->getPaginator($queryBuilder);
     }
 
-    protected function getPaginator(QueryBuilder $queryBuilder): Pagerfanta
+    // Use output walkers option in DoctrineORMAdapter should be false as it affects performance greatly (see #3775)
+    protected function getPaginator(QueryBuilder $queryBuilder, $fetchJoinCollection = true, $useOutputWalker = false): Pagerfanta
```
lchrusciel: It is a BC-Break and it cannot be merged. We have to preserve current behavior (also for following reasons: https://github.com/Sylius/Sylius/issues/3775).

This could be fixed with the possibility to of changing this behavior with some config perhaps or overriding it on your own.

tom10271: Understand it might be BC-Break not it is clearly a bug. I will provide a case to make things clear. Resources repository setup via SyliusResourceBundle are all affected. I have saw many users asking why number of rows returned is less than sepcified.

https://github.com/Sylius/Sylius/pull/8038
https://github.com/Sylius/Sylius/pull/6674

And a few more but browsing histories are left in another computer I use to work.

lchrusciel: I could be a little bit harsh, with the first message. Excuse me for that. 

Thanks for posting a PR 👍

For sure returning a wrong amount of items on pagination is a bug and we should do something about it. Nonetheless, as far as I'm concerned about these options, they have a major performance impact and may break someone's website (with larger data sets).

These configuration options should stay as they are (unless 2.0 is planned to be released) and the proper solution would be to allow to easily configure them for the end-users. With the current behavior as default in the first iteration.

/cc @Sylius/core-team

lchrusciel: If there were not any better ideas, I would recommend allowing changing these fields in request configuration, so by default, you will have old logic, but end users will be able to adjust this config easily.

tom10271: Do you mean method signature for `request configuration`? I am fine to set it as false for $fetchJoinCollection to at least keep everything the same for now and let me fix things I need.

Thanks

lchrusciel: We would need to think about it, but I suppose it could be part of the repository configuration described here: https://docs.sylius.com/en/latest/components_and_bundles/bundles/SyliusResourceBundle/index_resources.html#using-a-custom-repository-method

Zales0123:
I think it can be closed, as it was fixed in the meantime 🖖 Nevertheless, thank you, @tom10271 for your contribution!

PR #160, opened 2020-04-27
ResourceBundle documentation extracted to its repository

SirDomin:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

### DIFF
--- a/docs/index_resources.md
+++ b/docs/index_resources.md
@@ -4,11 +4,13 @@ To get a paginated list of Books, we will use **indexAction** of our controller.
 In the default scenario, it will return an instance of paginator, with a list of Books.
 
 ```yaml
-    app_book_index:
-        path: /books
-        methods: [GET]
-        defaults:
-            _controller: app.controller.book:indexAction
+# config/routes.yaml
+
+app_book_index:
+    path: /books
+    methods: [GET]
+    defaults:
+        _controller: app.controller.book:indexAction
 ```
 When you go to ``/books``, the ResourceController will use the repository (``app.repository.book``) to create a paginator.
 The default template will be rendered - ``App:Book:index.html.twig`` with the paginator as the ``books`` variable.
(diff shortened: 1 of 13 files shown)


docs/index_resources.md
```diff
@@ -0,0 +1,138 @@
...
+It will transform your doctrine query builder into ``Pagerfanta\Pagerfanta`` object.
+
+## Changing the "Max Per Page" Option of Paginator
+
+You can also control the "max per page" for paginator, using ``paginate`` parameter.
+
+```yaml
+    # config/routes.yaml
+
+    app_book_index_top:
```
lchrusciel, line 78: I would say, that YAML files shouldn't be indented

lchrusciel: That's true for all blocks (bash, yaml, PHP)

mamazu: Agreed. It makes pasting the code a little harder.

docs/routing.md
```diff
@@ -0,0 +1,289 @@
...
+
+```yaml
+    app_book:
+        resource: |
+            alias: app.book
+        type: sylius.resource
+```
+Results in the following routes:
+
+```bash
```
mamazu, line 17: This is technically not correct. Only the `php bin/console debug:router` is bash. The rest is only text output where no syntax highlighting should be used.
Or maybe we just use the table function of markdown?

docs/update_resource.md
```diff
@@ -0,0 +1,197 @@
...
+                event: customer_update
+```
+This way, you can listen to ``app.book.pre_customer_update`` and ``app.book.post_customer_update`` events. It's especially useful, when you use
+``ResourceController:updateAction`` in more than one route.
+
+
+## [API] Returning resource or no content
+
+Depending on your app approach it can be useful to return a changed object or only the ``204 HTTP Code``, which indicates that everything worked smoothly.
+Sylius, by default is returning the ``204 HTTP Code``, which indicates an empty response. If you would like to receive a whole object as a response you should set a `return_content` option to true.
```
mamazu, line 152: This sentence only uses a single \` here although 2 would be consistent with the rest of the documentation. (Although one \` would have been enough.

lchrusciel:
Thanks, @SirDomin! :tada:

PR #167, opened 2020-06-04
Upgrade rest bundle

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | not anymore
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/UPGRADE-1.7.md
+++ b/UPGRADE-1.7.md
@@ -8,5 +8,5 @@ fos_rest:
         enabled: true 
 ```
 
-* **BC BREAK**: `ViewHandler` constructor now requires `FOS\RestBundle\View\ConfigurableViewHandlerInterface` as first argument instead of 
+* **BC BREAK**: `Sylius\Bundle\ResourceBundle\Controller\ViewHandler` constructor now requires `FOS\RestBundle\View\ConfigurableViewHandlerInterface` as first argument instead of 
     `FOS\RestBundle\View\ViewHandler`.
--- a/UPGRADE-1.7.md
+++ b/UPGRADE-1.7.md
@@ -1,4 +1,0 @@
-# UPGRADE FROM `v1.6.*` TO `v1.7.0`
-
-* **BC BREAK**: `Sylius\Bundle\ResourceBundle\Controller\ViewHandler` constructor now requires `FOS\RestBundle\View\ConfigurableViewHandlerInterface` as first argument instead of 
-    `FOS\RestBundle\View\ViewHandler`.
--- a/UPGRADE-1.7.md
+++ b/UPGRADE-1.7.md
@@ -1,12 +1,4 @@
 # UPGRADE FROM `v1.6.*` TO `v1.7.0`
 
-* **BC BREAK**: Edit your `fos_rest` config to enable the body listener as follows:
-```yaml
-fos_rest:
-    # [...]
-    body_listener:
-        enabled: true 
-```
-
 * **BC BREAK**: `Sylius\Bundle\ResourceBundle\Controller\ViewHandler` constructor now requires `FOS\RestBundle\View\ConfigurableViewHandlerInterface` as first argument instead of 
     `FOS\RestBundle\View\ViewHandler`.
--- a/UPGRADE-1.7.md
+++ b/UPGRADE-1.7.md
@@ -1,0 +1,12 @@
+# UPGRADE FROM `v1.6.*` TO `v1.7.0`
+
+* **BC BREAK**: Edit your fos_rest config to enable the body listener as following:
+```yaml
+fos_rest:
+    # [...]
+    body_listener:
+        enabled: true 
+```
+
+* **BC BREAK**: `ViewHandler` constructor now requires `FOS\RestBundle\View\ConfigurableViewHandlerInterface` as first argument instead of 
+    `FOS\RestBundle\View\ViewHandler`.
--- a/UPGRADE-1.7.md
+++ b/UPGRADE-1.7.md
@@ -1,6 +1,6 @@
 # UPGRADE FROM `v1.6.*` TO `v1.7.0`
 
-* **BC BREAK**: Edit your fos_rest config to enable the body listener as following:
+* **BC BREAK**: Edit your `fos_rest` config to enable the body listener as follows:
 ```yaml
 fos_rest:
     # [...]
--- a/src/Bundle/Controller/Parameters.php
+++ b/src/Bundle/Controller/Parameters.php
@@ -19,8 +19,10 @@ class Parameters extends ParameterBag
 {
     /**
      * {@inheritdoc}
+     *
+     * @psalm-param string $path
      */
-    public function get(string $path, $default = null)
+    public function get($path, $default = null)
     {
         $result = parent::get($path, $default);
 
--- a/src/Bundle/Controller/Parameters.php
+++ b/src/Bundle/Controller/Parameters.php
@@ -20,7 +20,7 @@ class Parameters extends ParameterBag
     /**
      * {@inheritdoc}
      */
-    public function get($path, $default = null)
+    public function get(string $path, $default = null)
     {
         $result = parent::get($path, $default);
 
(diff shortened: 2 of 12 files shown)


loic425:
@pamil we have a lot of work here :rofl:

loic425:
@lchrusciel @pamil 
I worked a lot today on this pull request. Tell me if I am in the good direction or not.

src/Bundle/Controller/Parameters.php
```diff
@@ -20,7 +20,7 @@ class Parameters extends ParameterBag
     /**
      * {@inheritdoc}
      */
-    public function get($path, $default = null)
+    public function get(string $path, $default = null)
```
pamil, line 23: Is it required? It's a BC break unfortunately.

src/Bundle/Controller/ViewHandler.php
```diff
@@ -13,16 +13,16 @@
...
 use Symfony\Component\HttpFoundation\Response;
 
 final class ViewHandler implements ViewHandlerInterface
 {
-    /** @var RestViewHandler */
+    /** @var ConfigurableViewHandlerInterface */
     private $restViewHandler;
 
-    public function __construct(RestViewHandler $restViewHandler)
+    public function __construct(ConfigurableViewHandlerInterface $restViewHandler)
```
pamil, line 25: It's a slight BC break, but since this PR bumps up the requirement for FOSRestBundle I don't think there's any case one would pass RestViewHandler to the constructor.

src/Bundle/Form/Extension/HttpFoundation/HttpFoundationRequestHandler.php
```diff
@@ -89,6 +89,7 @@ public function handleRequest(FormInterface $form, $request = null)
                 $params = $request->request->all();
                 $files = $request->files->all();
             } elseif ($request->request->has($name) || $request->files->has($name)) {
+                /** @psalm-var mixed */
```
pamil, line 92:
```suggestion
                /** @psalm-var mixed $default */
```

src/Bundle/test/app/config/config.yml
```diff
@@ -32,7 +32,11 @@ doctrine:
                 auto_mapping: true
 
 fos_rest:
+    body_listener:
+        enabled: true
     view:
+        view_response_listener:
+            enabled: true
```
pamil, line 39: Are those config changes required? If so, we should mention them in UPGRADE.md.

src/Bundle/Controller/Parameters.php
loic425: for psalm only, but I can suppress this psalm warning.

src/Bundle/Controller/ViewHandler.php
loic425: Yes and it's a final class. So we can't override it.

src/Bundle/test/app/config/config.yml
loic425: Yes it's required. I have to check if both are required.

src/Bundle/Controller/ViewHandler.php
loic425: @pamil I am wondering why is it a bc break? Cause this service is marked as final and the old RestViewHandler argument implements ConfigurableViewHandlerInterface.
If we inject the old one with dependency injection, it continues to work.

UPGRADE-1.7.md
```diff
@@ -0,0 +1,12 @@
...
+
+* **BC BREAK**: Edit your fos_rest config to enable the body listener as following:
+```yaml
+fos_rest:
+    # [...]
+    body_listener:
+        enabled: true 
+```
+
+* **BC BREAK**: `ViewHandler` constructor now requires `FOS\RestBundle\View\ConfigurableViewHandlerInterface` as first argument instead of
```
pamil, line 11:
```suggestion
* **BC BREAK**: `Sylius\Bundle\ResourceBundle\Controller\ViewHandler` constructor now requires `FOS\RestBundle\View\ConfigurableViewHandlerInterface` as first argument instead of 
```

UPGRADE-1.7.md
```diff
@@ -0,0 +1,12 @@
+# UPGRADE FROM `v1.6.*` TO `v1.7.0`
+
+* **BC BREAK**: Edit your fos_rest config to enable the body listener as following:
```
pamil, line 3:
```suggestion
* **BC BREAK**: Edit your `fos_rest` config to enable the body listener as follows:
```

src/Bundle/Controller/ViewHandler.php
pamil: Theoretically, someone could create their own service definition where they inject the original service into it.

src/Bundle/Form/Extension/HttpFoundation/HttpFoundationRequestHandler.php
```diff
@@ -89,6 +89,7 @@ public function handleRequest(FormInterface $form, $request = null)
                 $params = $request->request->all();
                 $files = $request->files->all();
             } elseif ($request->request->has($name) || $request->files->has($name)) {
+                /** @psalm-var mixed $default */
```
pamil, line 92:
```suggestion
                /** @psalm-var array|null $default */
```

src/Bundle/test/app/config/config.yml
```diff
@@ -32,6 +32,8 @@ doctrine:
                 auto_mapping: true
 
 fos_rest:
+    body_listener:
```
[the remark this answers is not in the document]
pamil: I wonder whether we could do that automatically - let's say by prepending the config using `PrependExtensionInterface`?

src/Bundle/Controller/ViewHandler.php
loic425: Yes but IMHO it should work too, cause I replaced with the implemented interface. Inject the old service definition should work too.

src/Bundle/test/app/config/config.yml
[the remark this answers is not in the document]
loic425: Yes you're right, it's a good idea.

src/Bundle/Controller/ViewHandler.php
pamil: Ah, that's perfect then! Sorry, I must've missed it.

loic425: Do we still need this mention in UPGRADE doc?

src/Bundle/test/app/config/config.yml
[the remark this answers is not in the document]
loic425: @pamil It works!!! 🎉

src/Bundle/Controller/ViewHandler.php
pamil: No need then :)

src/Bundle/Controller/ResourcesCollectionProvider.php
```diff
@@ -58,7 +58,8 @@ public function get(RequestConfiguration $requestConfiguration, RepositoryInterf
                 $requestConfiguration->getPaginationMaxPerPage(),
                 $paginationLimits
             ));
-            $paginator->setCurrentPage($request->query->get('page', 1));
+            $currentPage = (int) $request->query->get('page', '1');
```
lchrusciel, line 61: Woulndn't it be better?
```suggestion
            $currentPage = (int) $request->query->getInt('page', '1');
```

loic425: @lchrusciel Yes, but, if it returns an integer, we don't need the (int) conversion after :)

loic425, line 61: @lchrusciel 
```suggestion
            $currentPage = $request->query->getInt('page', '1');
```

loic425: I pushed another suggestion to remote the unnecessary int conversion.

pamil:
Thank you, Loïc! :tada:

PR #173, opened 2020-07-05
[WIP] Remove rest dependencies

### LINKED ISSUE #170: Remove FosRestBundle and BazingaHateoasBundle dependency.
# Reason 1
The two bundles are very usefull to build RESTfull API via HAL,   I also like to support [JSONAPI](https://jsonapi.org/), so I suggest move they into a new package, maybe named as HAL extension for SyliusResourceBundle. 

# Reason 2

I'd like transform my entity to json using [fractal](https://github.com/thephpleague/fractal)  instead of jmsserializer.

even though jmsserializer is flexiable and powerful , but it makes my API not stable enough, because they share the same metadata, for example, someone adds a  'roles'  property to my user class within 'Default' groups.

```
Acme\Component\User\Model\User:
    exclusion_policy: ALL
    xml_root_name: xiaolvmo_user
    properties:
        id:
            expose: true
            type: integer
            groups: [Default]
        username:
            expose: true
            type: string
            groups: [Default]
        roles:
            expose: true
            type: ArrayCollection<Acme\Component\User\Model\Role>
            groups: [Default]

```

if Order also has user relation,  then it will output roles, which may be not what need.   serializing the roles property  may trigger a doctrine/orm lazy loading, which hurt performance. so the API is very fragile because of jmsserializer.

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #170
| License         | MIT


src/Bundle/DependencyInjection/Compiler/RegisterViewHandlerPass.php
```diff
@@ -0,0 +1,32 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+class RegisterViewHandlerPass implements CompilerPassInterface
```
lchrusciel, line 20: Could be finialized probably

loic425:
I think we have to add these librairies on conflict to prevent installing older versions than compatible ones. See Symfony composer requirements...

loic425:
I'd like to test prod requirements too. but I don't manage to launch clear cache command on --env=prod after composer install --no-dev. Test app still have test env...

loic425:
@pamil @lchrusciel I need help to test prod requirements. It doesn't work with my last commit.

lchrusciel:
Hey Loic, sorry for not being to responsive. We had a few commitments that we had to fulfill recently. However, with 1.8 release coming, you can expect more power to review and push things further. 

In the same time, thanks a lot for your patience and work :)

loic425:
@pamil @lchrusciel I rebased this PR to fix conflicts. I can squash all commits into one if you want.

lchrusciel:
Hey Loic,

thanks for your contribution and again sorry for the long feedback loop. However, we have to fix the main build before we can go anywhere further.

loic425:
@lchrusciel I fixed the Ci build.

loic425:
@lchrusciel @pamil This one is ready and build is fixed. Is there anything else to improve?

loic425:
@lchrusciel I made a little change to use `on-invalid="null"`, it allows to remove a compiler pass just for that case.

src/Bundle/DependencyInjection/Driver/AbstractDriver.php
```diff
@@ -60,13 +61,15 @@ protected function setClassesParameters(ContainerBuilder $container, MetadataInt
 
     protected function addController(ContainerBuilder $container, MetadataInterface $metadata): void
     {
+        $viewHandler = class_exists(FOSRestBundle::class) ? new Reference('sylius.resource_controller.view_handler') : null;
```
pamil, line 64:
```suggestion
        $viewHandler = new Reference('sylius.resource_controller.view_handler', ContainerInterface::NULL_ON_INVALID_REFERENCE);
```

src/Bundle/Resources/config/services/controller.xml
```diff
@@ -48,7 +48,7 @@
             <argument type="service" id="event_dispatcher" />
         </service>
         <service id="sylius.resource_controller.view_handler" class="Sylius\Bundle\ResourceBundle\Controller\ViewHandler" public="false">
-            <argument type="service" id="fos_rest.view_handler" />
+            <argument type="service" id="fos_rest.view_handler" on-invalid="null"/>
```
pamil, line 51: Just a coding standard :)

```suggestion
            <argument type="service" id="fos_rest.view_handler" on-invalid="null" />
```

src/Bundle/test/app/AppKernel.php
```diff
@@ -22,29 +22,40 @@ class AppKernel extends Kernel
      */
     public function registerBundles()
     {
-        return [
+        if ('prod' === $this->getEnvironment()) {
+            $loader = require __DIR__ . '/../../../../vendor/autoload.php';
+            $loader->addPsr4('AppBundle\\', __DIR__ . '/../src/AppBundle/');
+            $loader->addPsr4('Webmozart\\Assert\\', __DIR__ . '/../../../Component/vendor/webmozart/assert/src/');
```
pamil, line 28: I guess it's related to #188 and it should be fixed.

loic425: @pamil Maybe, we could fix #188 and rebase this one (again).

pamil: Done in #189 🎉

loic425: @pamil I rebased (again)

composer.json
```diff
@@ -43,18 +41,22 @@
...
-        "willdurand/hateoas-bundle": "^2.0",
         "winzou/state-machine-bundle": "^0.3.2 || ^0.4.3 || ^0.5"
     },
     "replace": {
         "sylius/resource": "self.version"
     },
     "conflict": {
-        "amphp/amp": "^2.4.3"
+        "amphp/amp": "^2.4.3",
+        "friendsofsymfony/rest-bundle": "<3.0",
```
pamil, line 51: I wonder whether we shouldn't also add conflict for the next major version (to be removed when we decide to support it). Then it would be `<3.0 >=4.0` (and similar for other requirements here). What do you think?

src/Bundle/DependencyInjection/Driver/AbstractDriver.php
```diff
@@ -60,13 +61,15 @@ protected function setClassesParameters(ContainerBuilder $container, MetadataInt
...
+        $viewHandler = new Reference('sylius.resource_controller.view_handler', ContainerInterface::NULL_ON_INVALID_REFERENCE);
+
         $definition = new Definition($metadata->getClass('controller'));
         $definition
             ->setPublic(true)
             ->setArguments([
                 $this->getMetadataDefinition($metadata),
                 new Reference('sylius.resource_controller.request_configuration_factory'),
-                new Reference('sylius.resource_controller.view_handler'),
+                $viewHandler,
```
pamil, line 72:
```suggestion
                new Reference('sylius.resource_controller.view_handler', ContainerInterface::NULL_ON_INVALID_REFERENCE),
```

I guess we can bring it inline now, so the only change would be adding the second argument to `Reference` constructor.

src/Bundle/test/app/AppKernel.php
```diff
@@ -22,29 +22,39 @@ class AppKernel extends Kernel
      */
     public function registerBundles()
     {
-        return [
+        if ('prod' === $this->getEnvironment()) {
+            $loader = require __DIR__ . '/../../../../vendor/autoload.php';
+            $loader->addPsr4('AppBundle\\', __DIR__ . '/../src/AppBundle/');
```
pamil, line 27: Is it needed? `autoload-dev` in `composer.json` should already do the job.

src/Bundle/test/app/AppKernel.php
```diff
@@ -22,29 +22,39 @@ class AppKernel extends Kernel
...
             new Symfony\Bundle\TwigBundle\TwigBundle(),
-            new \winzou\Bundle\StateMachineBundle\winzouStateMachineBundle(),
+            new winzou\Bundle\StateMachineBundle\winzouStateMachineBundle(),
             new Stof\DoctrineExtensionsBundle\StofDoctrineExtensionsBundle(),
-            new Fidry\AliceDataFixtures\Bridge\Symfony\FidryAliceDataFixturesBundle(),
-            new Nelmio\Alice\Bridge\Symfony\NelmioAliceBundle(),
             new AppBundle\AppBundle(),
         ];
+
+        if (in_array($this->getEnvironment(), ['dev', 'test'])) {
```
pamil, line 41: Why don't we always add those bundles? They're always included when installed with `require-dev` dependencies.

src/Bundle/test/app/console
```diff
@@ -19,6 +19,7 @@ use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Component\Console\Input\ArgvInput;
 
 $input = new ArgvInput();
+$env = $input->getParameterOption(array('--env', '-e'), getenv('SYMFONY_ENV') ?: 'test');
 
-$application = new Application(new AppKernel('test', true));
+$application = new Application(new AppKernel($env, true));
```
pamil, line 24: What's the idea behind that? Do we really need it?

src/Bundle/test/app/AppKernel.php
loic425: Nope, it's removed on "prod" env.

src/Bundle/test/app/console
loic425: The idea is to test prod requirements (without optional rest bundles).

loic425: @pamil Maybe you have a better idea to test without optional bundles...

src/Bundle/test/app/AppKernel.php
pamil: When do we run the test application in `prod` environment?

src/Bundle/test/app/AppKernel.php
pamil:
```suggestion
        if (in_array($this->getEnvironment(), ['dev', 'test'], true)) {
```

.travis.yml
```diff
@@ -62,3 +62,7 @@ script:
     - (cd src/Component && vendor/bin/phpspec run)
 
     - vendor/bin/phpunit
+
+    - composer require doctrine/orm:^2.5 --no-update --no-scripts --prefer-dist
+    - composer update --no-dev --prefer-dist
+    - (cd src/Bundle/test && app/console cache:clear --env=prod)
```
loic425, line 68: @pamil we test prod environment here.

src/Bundle/test/app/AppKernel.php
```diff
@@ -49,23 +40,9 @@ public function registerBundles()
...
-     */
-    protected function getContainerBaseClass()
-    {
-        if (0 === strpos($this->environment, 'test')) {
-            return MockerContainer::class;
-        }
-
-        return parent::getContainerBaseClass();
+        $loader->load(__DIR__ . '/config/{config}.{php,xml,yaml,yml}', 'glob');
+        $loader->load(__DIR__ . '/config/{config}_' . $this->getEnvironment() . '.{php,xml,yaml,yml}', 'glob');
```
pamil, line 46: Makes the environment config file optional, it's the same behaviour as in Symfony 4+ apps.

src/Bundle/test/app/console
```diff
@@ -19,7 +19,7 @@ use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Component\Console\Input\ArgvInput;
 
 $input = new ArgvInput();
-$env = $input->getParameterOption(array('--env', '-e'), getenv('SYMFONY_ENV') ?: 'test');
+$env = $input->getParameterOption(['--env', '-e'], getenv('APP_ENV') ?? 'test');
```
pamil, line 22: Replaced `SYMFONY_ENV` to more recent `APP_ENV`.

pamil:

PR #169, opened 2020-06-11
Upgrade pagerfanta

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

PR #175, opened 2020-07-10
Upgrade PagerfantaBundle to new version with B/C layer

mbabker:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | yes
| Related tickets | https://github.com/Sylius/Sylius/issues/10928
| License         | MIT

This is an alternative version of https://github.com/Sylius/SyliusResourceBundle/pull/169 with a B/C layer to try and map the bundle configuration from `WhiteOctoberPagerfantaBundle` to `BabDevPagerfantaBundle`.

The B/C layer involves interfacing directly with the container in a few ways:

- Using a custom container extension that can process the configuration under `white_october_pagerfanta` and map it to the `babdev_pagerfanta` configuration (since none of the child nodes were renamed, this part is pretty simple) and re-create the renamed `white_october_pagerfanta.default_view` parameter
- Using a compiler pass to handle the `white_october_pagerfanta.view_factory.class` parameter that I fully dropped to change the `Pagerfanta\View\ViewFactory` class to something else if so desired and to alias the renamed `twig.extension.pagerfanta` and `white_october_pagerfanta.view_factory` service IDs (the aliases are deprecated in favor of the new bundle's service IDs)

The one part of https://github.com/Sylius/SyliusResourceBundle/pull/169#issuecomment-644839570 not handled is a class mapping layer for the namespace change.  It'd be easy to throw together a file with a bunch of `class_alias()` calls, but as also noted in the comment, a bunch of classes were made final so if someone were extending them then even the aliasing wouldn't save them from that API break.

In addition to just upgrading to a supported bundle version that also removes one roadblock to full Symfony 5 support, there are a number of new features in the Pagerfanta library and/or PagerfantaBundle that could be interesting to Sylius users, including:

- Customizable route generation logic (no longer hardcoded into the Twig extension)
- Twig template support alongside the pre-existing PHP templates
- Tailwind CSS support (I've shipped the Twig template for it, need to add the PHP version still)
- At some point serializer support will land (at work we're manually building a lot of responses from our API with data out of the `Pagerfanta\Pagerfanta` class, at some point it's going to make sense to move that logic into the serializers (both Symfony's and JMS') and improve the pager's use in an API context)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ install:
             symfony/form:$SYMFONY_VERSION \
             symfony/framework-bundle:$SYMFONY_VERSION \
             symfony/security-csrf:$SYMFONY_VERSION \
+            symfony/translation:$SYMFONY_VERSION \
             symfony/twig-bundle:$SYMFONY_VERSION \
             symfony/validator:$SYMFONY_VERSION \
             symfony/yaml:$SYMFONY_VERSION \
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,7 @@
     "require": {
         "php": "^7.2",
 
+        "babdev/pagerfanta-bundle": "^2.0",
         "doctrine/doctrine-bundle": "^1.12|^2.0",
         "doctrine/persistence": "^1.3",
         "friendsofsymfony/rest-bundle": "^3.0",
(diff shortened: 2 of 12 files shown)


stloyd:
@lchrusciel @pamil Is there anything blocking this? We would love to use Sylius + Symfony 5 but we're still locked due too few things :(

lchrusciel:
Hey Joseph, we are blocked mostly by our current commitments. We will enter a feature freeze next week, so probably we will put more effort into Symfony5 compatibility since then.

lchrusciel:
Thank you, Michael! :tada:

PR #177, opened 2020-08-31
Pagerfanta updates

mbabker:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

As noted [here](https://www.babdev.com/open-source/updates/pagerfanta-package-structure-updates), the latest version of the Pagerfanta package was restructured to support separately installed sub-packages.  So, to help with avoiding deprecation notices and preparing for compatibility with the new major version I expect to push out toward the end of the year, I've bumped the Pagerfanta minimums up to the latest versions, updated the query adapter classes to use the non-deprecated classes, and changed the component to install only the `pagerfanta/core` sub-package since that's all it directly depends on.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,8 +26,8 @@
     ],
     "require": {
         "php": "^7.3",
-        "babdev/pagerfanta-bundle": "^2.0",
-        "doctrine/doctrine-bundle": "^1.12 || ^2.0",
+        "babdev/pagerfanta-bundle": "^2.5",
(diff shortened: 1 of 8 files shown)


lchrusciel:
We have to fix the main build before we can go anywhere further.

pamil (approved):

pamil:
Thank you, Michael! :tada:

PR #178, opened 2020-09-01
When using winzouStateMachineBundle 0.4, the old named services are aliases, so need to be marked public as well

mbabker:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

In winzouStateMachineBundle 0.4, the `sm.factory`, `sm.callback_factory`, and `sm.callback.cascade_transition` IDs are aliases and not real services.  So, the aliases need to be made public as well otherwise they won't be available through the container once it's compiled.

### DIFF
--- a/src/Bundle/DependencyInjection/Compiler/WinzouStateMachinePass.php
+++ b/src/Bundle/DependencyInjection/Compiler/WinzouStateMachinePass.php
@@ -28,9 +28,6 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
  */
 final class WinzouStateMachinePass implements CompilerPassInterface
 {
-    /**
-     * {@inheritdoc}
-     */
     public function process(ContainerBuilder $container): void
     {
         if ($container->hasDefinition('sm.factory') && !$container->hasDefinition(FactoryInterface::class)) {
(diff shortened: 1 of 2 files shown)


lchrusciel (approved):

stloyd (approved):

src/Bundle/DependencyInjection/Compiler/WinzouStateMachinePass.php
```diff
@@ -39,9 +38,12 @@ public function process(ContainerBuilder $container): void
         $container->setAlias('sm.callback.cascade_transition', CascadeTransitionCallback::class);
```
stloyd, line 38: IMHO this should be changed (tested the current proposal & it didn't work for me).

```suggestion
        $container->setAlias('sm.factory', FactoryInterface::class)->setPublic(true);
        $container->setAlias('sm.callback_factory', CallbackFactoryInterface::class)->setPublic(true);
        $container->setAlias('sm.callback.cascade_transition', CascadeTransitionCallback::class)->setPublic(true);
```

stloyd (changes requested):

src/Bundle/DependencyInjection/Compiler/WinzouStateMachinePass.php
mbabker: PR seems fine as is for me, this is coming out of my local Sylius clone with my WIP upgrade PR applied:

```sh
Michaels-MacBook-Pro-2:Sylius mbabker$ bin/console debug:container sm.factory

 // This service is a public alias for the service SM\Factory\Factory
```

Compared to the current state of affairs:

```sh
Michaels-MacBook-Pro-2:Sylius mbabker$ bin/console debug:container sm.factory

 // This service is a private alias for the service SM\Factory\Factory                                                  

... snip ...

 ! [NOTE] The "sm.factory" service or alias has been removed or inlined when the container was compiled.
```

lchrusciel:
We have to fix the main build before we can go anywhere further.

pamil:
Rebased & added a functional test for these services accessibility.

pamil:
Thank you, Michael! :1st_place_medal:

PR #181, opened 2020-09-08
Fix build

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         "symfony/validator": "^4.4|^5.0",
         "symfony/yaml": "^4.4|^5.0",
         "willdurand/hateoas-bundle": "^2.0",
-        "winzou/state-machine-bundle": "^0.3|^0.4"
+        "winzou/state-machine-bundle": "^0.4.3"
     },
     "require-dev": {
         "doctrine/orm": "^2.5",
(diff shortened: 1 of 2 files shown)


loic425:
@lchrusciel What do you think about these fixes?

composer.json
```diff
@@ -39,7 +39,7 @@
         "symfony/validator": "^4.4|^5.0",
         "symfony/yaml": "^4.4|^5.0",
         "willdurand/hateoas-bundle": "^2.0",
-        "winzou/state-machine-bundle": "^0.3|^0.4"
+        "winzou/state-machine-bundle": "^0.4.3"
```
lchrusciel, line 42: Why is it needed?

lchrusciel (commented):
Hey Loic,

thanks for your work and fixes seem ok. However, I'm wondering why do we need to bump the state machine version? Not a big deal, but we are using `0.3` in Sylius right now.

composer.json
loic425: There's a bug on winzou with dependencies between state-machine and state-machine bundle. Maybe there's something better to do:
"winzou/state-machine-bundle": "^0.3.2|^0.4.3"
I'm gonna try that.

loic425: @lchrusciel But I think we have to improve CI configuration to test these two state machine versions.

lchrusciel (approved):

lchrusciel:
Thanks, Loïc! :1st_place_medal:

PR #182, opened 2020-09-09
Testing several state machine versions

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | allows update to Symfony 5
| License         | MIT

This pull request replaces this one #174 and add tests for all state machine versions.
It's also for Symfony 5 support: https://github.com/Sylius/SyliusResourceBundle/issues/121

PR #186, opened 2020-09-29
Fix Winzou sm.factory with Bundle 0.3

macintoshplus:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

When this master version is used with Winzou State Machine Bundle 0.3 and Sylius dev-master, the dependency injection throw this exception:

```
[Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]                    
  The service "twig" has a dependency on a non-existent service "SM\Factory\FactoryInterface".
```

It's true because the service id is `sm.factory` and a alias is defined with same name.

The realy behavior need is define alias from `sm.factory` to interface if the service id `sm.factory` exists. And define alias fromt interface to `sm.factory` if service id `sm.factory` is not defined.

And same behavior for service `sm.callback_factory` and `sm.callback.cascade_transition`.

### DIFF
--- a/src/Bundle/DependencyInjection/Compiler/WinzouStateMachinePass.php
+++ b/src/Bundle/DependencyInjection/Compiler/WinzouStateMachinePass.php
@@ -34,9 +34,21 @@ final class WinzouStateMachinePass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container): void
     {
-        $container->setAlias('sm.factory', FactoryInterface::class);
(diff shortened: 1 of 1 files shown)


loic425:
@macintoshplus I fixed that in #182

macintoshplus:
@loic425 Nice and good job. I close my PR.

PR #200, opened 2020-10-14
Use HTTPS instead of HTTP in links in composer.json

ValentineJester:


pamil:
Thanks, @ValentineJester! :tada:

PR #204, opened 2020-10-15
Replace AbstractController with ControllerTrait & ContainerAwareInterface

pamil:
We can't use `AbstractController` if we want to keep BC - `AbstractController` requires a service locator with predefined dependencies and `ResourceController` requires a container with all public services for BC purposes.

### DIFF
--- a/src/Bundle/DependencyInjection/Driver/AbstractDriver.php
+++ b/src/Bundle/DependencyInjection/Driver/AbstractDriver.php
@@ -58,8 +58,6 @@ abstract class AbstractDriver implements DriverInterface
 
     protected function addController(ContainerBuilder $container, MetadataInterface $metadata): void
     {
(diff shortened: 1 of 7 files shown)


src/Bundle/DependencyInjection/Driver/AbstractDriver.php
```diff
@@ -58,8 +58,6 @@ protected function setClassesParameters(ContainerBuilder $container, MetadataInt
 
     protected function addController(ContainerBuilder $container, MetadataInterface $metadata): void
     {
-        $viewHandler = new Reference('sylius.resource_controller.view_handler', ContainerInterface::NULL_ON_INVALID_REFERENCE);
```
loic425, line 61: @pamil Why removing this?

PR #208, opened 2020-10-30
Use backslash for templates and form

### LINKED ISSUE #157: Error parsing 'templates' property in routing.yml for new Resource
This is roughly what I've been following and this specific section is what I was comparing my routing.yml to - http://docs.sylius.org/en/latest/bundles/SyliusResourceBundle/routing.html#using-custom-templates

The line `templates: AppTeamBundle:Team/Crud` seems to be getting parsed incorrectly. This code/config is from a Bundle which was setup on a prior version ~ 3 months ago and works correctly on the older project (1.0.0-beta.2) still.

When trying to install it on a new project (1.0.0-beta.3) I get an error:

    The file "alias: app_team.team_member
    section: admin
    templates: AppTeamBundle:Team" does not exist in /var/www/vhosts/example/src/App/TeamBundle/Resources/config/routing.yml (which is being imported from "/var/www/vhosts/example/app/config/routing.yml").

I tried a few bits but the actually fix is just swapping the `/` for a `\` so it becomes;
`templates: AppTeamBundle:Team\Crud`.

To me it feels like this is a bug with the parsing rather than a valid change to the setup which just hasn't been represented in the documentation. 

The routing config was originally like this;

    app_team_admin:
        resource: |
            alias: app_team.team_member
            section: admin
            templates: AppTeamBundle:Team/Crud
            except: ['show']
            redirect: update
            grid: app_team_admin
            vars:
                all:
                    subheader: app_team.ui.team_member # define a translation key for your entity subheader
                index:
                    icon: 'users' # choose an icon that will be displayed next to the subheader
        type: sylius.resource
        prefix: /admin

jacquesbh:
| Q               | A
| --------------- | -----
| Bug fix?        | yes, kind of :)
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #157
| License         | MIT

Using:
```yaml
templates: App/Project/Crud
```
would lead to an error:
```
[Application] Oct 30 14:49:50 |CRITICA| PHP    Uncaught Exception: The file "alias: harvest.harvest_project section: admin templates: Admin/Project" does not exist in /Users/jacques/Sites/operandi/apps/sylius/config/../src/Resources/config/app/routing/admin.yaml (which is being imported from "/Users/jacques/Sites/operandi/apps/sylius/config/routes.yaml").
```

I tried everything until I saw the use of backslash on another project… I tried and it works fine.

So, I've updated the documentation.

Also, the form is incorrect as well, but not related, I think since we use backslashes for PHP classes.

### DIFF
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -172,7 +172,7 @@ app_admin_book:
     resource: |
         alias: app.book
         section: admin
-        templates: Admin/Book
+        templates: Admin\Book
     type: sylius.resource
     prefix: /admin
 ```
@@ -191,7 +191,7 @@ If you want to use a custom form:
 app_book:
     resource: |
         alias: app.book
-        form: App/Form/Type/AdminBookType
+        form: App\Form\Type\AdminBookType
     type: sylius.resource
 ```
 ``create`` and ``update`` actions will use ``App/Form/Type/AdminBookType`` form type.


pamil:
Could we somehow add a test for that convention so that we can be sure it works? We've got a test application in `src/Bundle/test`.

loic425:
@jacquesbh @pamil To follow Symfony best practices, we should use lowercase for templates:
templates: admin\book

But this is a very old bug and I think we can fix it instead. I'm not a big fan to have a template path with backslashes.

jacquesbh:
> Could we somehow add a test for that convention so that we can be sure it works? We've got a test application in `src/Bundle/test`.

@pamil Yes. But it requires more time. So this PR will just wait for it :p.

> To follow Symfony best practices, we should use lowercase for templates:
templates: admin\book
> But this is a very old bug and I think we can fix it instead. I'm not a big fan to have a template path with backslashes.

@loic425 I agree with you it's an old bug. But fixing it will lead to BC breaks, won't it? Since it will break a lot of applications and bundles using the backslashes formatted paths.

pamil:
It shouldn't be needed after https://github.com/symfony/symfony/pull/39142 is released 🚀

loic425:
> It shouldn't be needed after [symfony/symfony#39142](https://github.com/symfony/symfony/pull/39142) is released 🚀

Great news!

PR #214, opened 2020-11-05
Do not use deprecated Doctrine Inflector API

pamil:


composer.json
```diff
@@ -27,7 +27,10 @@
     "require": {
         "php": "^7.3",
         "babdev/pagerfanta-bundle": "^2.5",
+        "doctrine/collections": "^1.6",
         "doctrine/doctrine-bundle": "^1.12 || ^2.0",
+        "doctrine/event-manager": "^1.1",
+        "doctrine/inflector": "^1.4 || ^2.0",
```
lchrusciel, line 33: This dependency will block newest resource bundle in the main Sylius repository: https://github.com/Sylius/Sylius/blob/master/composer.json#L195

We've conflicted with doctrine/inflector because of the fix of `taxon` word pluralization(from `taxons` to `taxa`). Proposed fixed may be found at: https://github.com/doctrine/inflector/issues/147

pamil: We should fix that in Sylius itself - `^1.4` is required because that's the first version that introduces forward compatibility with `^2.0`.

pamil: I'm just curious what's the best way to make that Inflector customisable - should we allow for `setInflector` static method on the metadata?

Issue #217, opened 2020-11-12
Duplicate initialisation of repositories


Fantus:
**Sylius version affected**: all

**Description**  
If you request a repository from the service container it will return a different repository instance than the one you will get from the `EntityManager` via `getRepository`

**Steps to reproduce**  

        $repository = $client->getContainer()->get('app.repository.book');

        $em = $client->getContainer()->get('app.manager.book');
        $repositoryAlias = $em->getRepository(Book::class);
        $this->assertTrue($repository === $repositoryAlias);

        $em = $client->getContainer()->get('doctrine.orm.entity_manager');
        $repositoryAlias = $em->getRepository(Book::class);
        $this->assertTrue($repository === $repositoryAlias);

![image](https://user-images.githubusercontent.com/1288316/98953616-30dc4580-24fd-11eb-9d35-706aad99b394.png)

**Possible Solution**  
https://github.com/Sylius/SyliusResourceBundle/pull/220

PR #218, opened 2020-11-12
Duplicate initialisation of repositories

Fantus:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | yes
| Related tickets | fixes https://github.com/Sylius/SyliusResourceBundle/issues/217
| License         | MIT

Accessing a repository via `EntityManger::getReposiory` and direct injection via the dot notation of the repository will yield different instances of the same repository. This is caused by:
* Repositories are not injected into the ServiceLocator of [ContainerRepositoryFactory](https://github.com/doctrine/DoctrineBundle/blob/1.12.x/Repository/ContainerRepositoryFactory.php) as the required tag `doctrine.repository_service` is missing for our generated the repositories
* Even when injected (after adding that `doctrine.repository_service`) the [ContainerRepositoryFactory::getRepository](https://github.com/doctrine/DoctrineBundle/blob/1.12.x/Repository/ContainerRepositoryFactory.php#L35) method cant resolve the repository as its injected with the dot notation and doctrine looks for the repository by class name
* Problem also exists for default repositories as tagging them is not enough as we cant switch to the class name as id

Solution/Changes;
For the solid repositories:
* **Tag the solid repositories** so they get injected
* **Switch to service id to class name**
* **Create a public alias** to maintain old functionality

For generic default repositories (cant be tagged as all use the same class)
* **Init default repositories via doctrine factory** so its using `ContainerRepositoryFactory` and by that prevent duplicate initialisation
* **Set default config value for default_repository_class to** `Sylius\Bundle\ResourceBundle\Doctrine\ORM\EntityRepository` (needed to prevent init of the standard `Doctrine\ORM\EntityRepository` in [ContainerRepositoryFactory::getOrCreateRepository](https://github.com/doctrine/DoctrineBundle/blob/1.12.x/Repository/ContainerRepositoryFactory.php#L77))

BC breaks:
- **Usage of dot notation in extension/compiler pass**: As mentioned in UPGRADE.md it breaks extension/bundles that do refer to repository via `$container->getDefinition` during the container compilation as `getDefinition` wont resolve an alias. This can be fixed by using `findDefinition` instead or using the new class name. The only place iam aware of that needs to be adjusted is the [SyliusChannelExtension](https://github.com/Sylius/Sylius/blob/7bdc5d406194da73becd749e48c57c9ae3fb95f4/src/Sylius/Bundle/ChannelBundle/DependencyInjection/SyliusChannelExtension.php#L40)

Deprecations:
- **Usage of dot notation:** I added a deprecation notice for the dot notation in preparation for the autowiring & autoconfiguration, but maybe this is to early and we could remove this again

I also needed to mute an error in phpstan and psaml about `Symfony\Component\DependencyInjection\Alias::setDeprecated` as the signature changed between symfony 4.4 and 5.x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## CHANGELOG FOR `1.7.x`
 
+#### Details
+- [#217](https://github.com/Sylius/SyliusResourceBundle/issues/202) Duplicate initialisation of repositories ([@Fantus](https://github.com/Sylius/SyliusResourceBundle/issues/217))
+
 ### v1.7.0-BETA.4 (2020-11-05)
 
 #### TL;DR
--- a/src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineORMDriver.php
+++ b/src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineORMDriver.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\ResourceBundle\DependencyInjection\Driver\Doctrine;
 
+use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\ServiceRepositoryCompilerPass;
 use Doctrine\Common\Persistence\ObjectManager as DeprecatedObjectManager;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
@@ -20,6 +21,7 @@ use Doctrine\Persistence\ObjectManager;
 use Sylius\Bundle\ResourceBundle\Doctrine\ORM\EntityRepository;
 use Sylius\Bundle\ResourceBundle\SyliusResourceBundle;
 use Sylius\Component\Resource\Metadata\MetadataInterface;
+use Symfony\Component\DependencyInjection\Alias;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## CHANGELOG FOR `1.7.x`
 
+#### Details
+- [#217](https://github.com/Sylius/SyliusResourceBundle/issues/202) Duplicate initialisation of repositories ([@Fantus](https://github.com/Sylius/SyliusResourceBundle/issues/217))
```
lchrusciel, line 4: Thanks for adding it, but we are generating this notation based on commits. Note just for future reference

src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineORMDriver.php
```diff
@@ -44,14 +46,34 @@ protected function addRepository(ContainerBuilder $container, MetadataInterface
...
+
+            $serviceAlias = new Alias($repositoryClass, true);
+            $usesSymfony51Api = method_exists(Alias::class, 'getDeprecation');
+            $deprecationMessage = sprintf('Using the service dot notation "%%alias_id%%" is deprecated. Use the class name "%s"', $repositoryClass);
+
+            if ($usesSymfony51Api) {
+                $serviceAlias->setDeprecated('sylius/resource-bundle', '1.7', $deprecationMessage);
+            } else {
+                $serviceAlias->setDeprecated(true, $deprecationMessage);
+            }
```
lchrusciel, line 72: As we discussed internally, I would prefer to skip the deprecation notice until we will agree on the expected new format of service declaration
```suggestion
```

src/Bundle/DependencyInjection/SyliusResourceExtension.php
```diff
@@ -73,6 +74,9 @@ public function prepend(ContainerBuilder $container): void
     {
         $config = ['body_listener' => ['enabled' => true]];
         $container->prependExtensionConfig('fos_rest', $config);
+
+        $config = ['orm' => ['default_repository_class' => EntityRepository::class]];
```
lchrusciel, line 78: Just a note for other reviewers. This line will by default set our `Sylius\Bundle\ResourceBundle\Doctrine\ORM\EntityRepository` as the default repository class as stated in the description. But it means that all non-resource-based doctrine entities will have two additional functions in their default repository class (add and remove), which could not be used, due to ResourceInterface typehint

lchrusciel (commented):
Hey Justus,

overall, thanks a lot for your contribution. For me, it solves more issues then generates, so I would be in favor of it, but I would like to hear another opinion about it as well. /cc @pamil

pamil (changes requested):

src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineORMDriver.php
```diff
@@ -44,14 +46,26 @@ protected function addRepository(ContainerBuilder $container, MetadataInterface
...
+            $container->setDefinition($aliasId, $definition);
+        } else {
+            // we cant use the doctrine factory as this will lead to an endless loop
+            $definition->setArguments([$managerReference, $this->getClassMetadataDefinition($metadata)]);
+            $definition->addTag(ServiceRepositoryCompilerPass::REPOSITORY_SERVICE_TAG);
+
+            $serviceAlias = new Alias($repositoryClass, true);
+
+            $container->setDefinition($repositoryClass, $definition);
+            $container->setAlias($aliasId, $serviceAlias);
```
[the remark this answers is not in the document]
Fantus: no sadly not, the alias gets not added to the `ServiceLocator` of the  ContainerRepositoryFactory. I tried that first and the lookup fails here: https://github.com/doctrine/DoctrineBundle/blob/85460b85edd8f61a16ad311e7ffc5d255d3c937c/Repository/ContainerRepositoryFactory.php#L43

src/Bundle/DependencyInjection/SyliusResourceExtension.php
Fantus: i made a version without the need to set this config. please check:
https://github.com/ecoco/SyliusResourceBundle/compare/%23217_duplicate_initialisation_of_repositories...ecoco:%23217_duplicate_initialisation_of_repositories_v2

and let me know what you think. If you like it should i just update this pull request or create a new one?

Fantus:
closed in favor of: https://github.com/Sylius/SyliusResourceBundle/pull/220

PR #220, opened 2020-11-22
Duplicate initialisation of repositories

Fantus:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/Sylius/SyliusResourceBundle/issues/217
| License         | MIT

Accessing a repository via `EntityManger::getReposiory` and direct injection via the dot notation of the repository will yield different instances of the same repository. This is caused by:
* Repositories are not injected into the ServiceLocator of [ContainerRepositoryFactory](https://github.com/doctrine/DoctrineBundle/blob/1.12.x/Repository/ContainerRepositoryFactory.php) as the required tag `doctrine.repository_service` is missing for our generated the repositories
* Even when injected (after adding that `doctrine.repository_service`) the [ContainerRepositoryFactory::getRepository](https://github.com/doctrine/DoctrineBundle/blob/1.12.x/Repository/ContainerRepositoryFactory.php#L35) method cant resolve the repository as its injected with the dot notation and doctrine looks for the repository by class name
* Problem also exists for default repositories but simply tagging them does not work

Solution/Changes;
For the solid repositories:
* **Tag the solid repositories** so they get injected
* **Switch to service id to class name**
* **Create a public alias** to maintain old functionality

For generic default repositories (cant be tagged as all use the same class)
* **Init default repositories via doctrine factory** so its using `ContainerRepositoryFactory` and by that prevent duplicate initialisation
* Decorated `ContainerRepositoryFactory` to be table to return the sylius `EntityRepository` instead of the default doctrine one

### DIFF
--- a/src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineORMDriver.php
+++ b/src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineORMDriver.php
@@ -21,7 +21,6 @@ use Doctrine\Persistence\ObjectManager;
 use Sylius\Bundle\ResourceBundle\Doctrine\ORM\EntityRepository;
 use Sylius\Bundle\ResourceBundle\SyliusResourceBundle;
 use Sylius\Component\Resource\Metadata\MetadataInterface;
-use Symfony\Component\DependencyInjection\Alias;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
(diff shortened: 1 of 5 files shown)


src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineORMDriver.php
```diff
@@ -44,14 +46,31 @@ protected function addRepository(ContainerBuilder $container, MetadataInterface
...
+            $repositoryFactoryDef->addMethodCall('addGenericEntity', [$entityClass]);
+        } else {
+            // we cant use the doctrine factory as this will lead to an endless loop
+            $definition->setArguments([$managerReference, $this->getClassMetadataDefinition($metadata)]);
+            $definition->addTag(ServiceRepositoryCompilerPass::REPOSITORY_SERVICE_TAG);
+
+            $serviceAlias = new Alias($repositoryClass, true);
+
+            $container->setDefinition($repositoryClass, $definition);
+            $container->setAlias($aliasId, $serviceAlias);
```
pamil, line 72: Would it remove the BC break?

```suggestion
            $definition->setArguments([$managerReference, $this->getClassMetadataDefinition($metadata)]);
            $container->setDefinition($aliasId, $definition);
            
            $doctrineDefinition = new Definition($repositoryClass);
            $doctrineDefinition->addTag(ServiceRepositoryCompilerPass::REPOSITORY_SERVICE_TAG);
            $doctrineDefinition->setFactory([new Reference('service_container', 'get']);
            $doctrineDefinition->setArguments([$aliasId]);
            $container->setDefinition($repositoryClass, $doctrineDefinition);
```

Fantus: adjusted accordingly, bc break is now gone

pamil (approved):

pamil:
Thank you, Justus! :tada:

PR #241, opened 2020-12-08
Test Symfony ^4.4 and ^5.1 instead of 4.4.* and 5.1.*

pamil:


loic425:
@pamil I think we have to add a "prefer-lowest" test

PR #255, opened 2021-01-14
Add autowire for resource Controllers

AdamKasp:
| Q               | A
| --------------- | -----
| Bug fix?        | no?
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/docs/create_resource.md
+++ b/docs/create_resource.md
@@ -225,4 +225,6 @@ app_genre_book_add:
                 parameters: { title: resource.title }
 ```
 
+Remember that you can use controller's Fully Qualified Class Name (``App\Controller\BookController``) instead of id ``app.controller.book`` 
+
 **[Go back to the documentation's index](index.md)**
--- a/src/Bundle/DependencyInjection/Compiler/RegisterControllerAliasesPass.php
+++ b/src/Bundle/DependencyInjection/Compiler/RegisterControllerAliasesPass.php
@@ -38,9 +38,10 @@ final class RegisterControllerAliasesPass implements CompilerPassInterface
             $controllerFQCN = $metadata->getClass('controller');
 
             if ($controllerFQCN !== ResourceController::class) {
-                $alias = new Alias($metadata->getServiceId('controller'));
-                $alias->setPublic(true);
-                $container->setAlias($controllerFQCN, $alias);
+                $definition = $container->getDefinition($metadata->getServiceId('controller'));
+
+                // TODO: Change to alias definition after bumping to > Symfony 5.2
+                $container->setDefinition($metadata->getClass('controller'), $definition);
             }
         }
     }
--- a/src/Bundle/DependencyInjection/Compiler/RegisterControllerAliasesPass.php
+++ b/src/Bundle/DependencyInjection/Compiler/RegisterControllerAliasesPass.php
@@ -1,0 +1,58 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler;
+
+use Sylius\Bundle\ResourceBundle\Controller\ResourceController;
+use Sylius\Component\Resource\Metadata\Metadata;
+use Sylius\Component\Resource\Model\ResourceInterface;
+use Symfony\Component\DependencyInjection\Alias;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
+
+final class RegisterControllerAliasesPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        try {
+            $resources = $container->getParameter('sylius.resources');
+        } catch (InvalidArgumentException $exception) {
+            return;
+        }
+
+        foreach ($resources as $alias => $configuration) {
+            $metadata = Metadata::fromAliasAndConfiguration($alias, $configuration);
+
+            $this->validateSyliusResource($metadata->getClass('model'));
+            $controllerFQCN = $metadata->getClass('controller');
+
+            if ($controllerFQCN !== ResourceController::class) {
+                $alias = new Alias($metadata->getServiceId('controller'));
+                $alias->setPublic(true);
+                $container->setAlias($controllerFQCN, $alias);
+            }
+        }
+    }
+
+    private function validateSyliusResource(string $class): void
+    {
+        if (!in_array(ResourceInterface::class, class_implements($class), true)) {
+            throw new InvalidArgumentException(sprintf(
+                'Class "%s" must implement "%s" to be registered as a Sylius resource.',
+                $class,
+                ResourceInterface::class
+            ));
+        }
+    }
+}
--- a/src/Bundle/Tests/DependencyInjection/Compiler/RegisterFqcnControllersPassTest.php
+++ b/src/Bundle/Tests/DependencyInjection/Compiler/RegisterFqcnControllersPassTest.php
@@ -15,29 +15,18 @@ namespace Sylius\Bundle\ResourceBundle\Tests\DependencyInjection\Compiler;
 
 use Matthias\SymfonyDependencyInjectionTest\PhpUnit\AbstractCompilerPassTestCase;
 use Sylius\Bundle\ResourceBundle\Controller\ResourceController;
-use Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler\RegisterControllerAliasesPass;
+use Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler\RegisterFqcnControllersPass;
 use Sylius\Component\Resource\Model\ResourceInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 
-final class RegisterControllerAliasesPassTest extends AbstractCompilerPassTestCase
+final class RegisterFqcnControllersPassTest extends AbstractCompilerPassTestCase
 {
     /**
      * @test
-     * @doesNotPerformAssertions
      */
-    public function it_throws_exception_if_resource_does_not_exist(): void
-    {
-        $this->compile();
-
-        $this->throwException(new InvalidArgumentException());
-    }
-
-    /**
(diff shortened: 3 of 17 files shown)


lchrusciel (commented):
`lint:container` command should be added to pipeline

loic425:
@lchrusciel @AdamKasp FYI, there is a hook here
https://github.com/Sylius/Sylius-Standard/blob/a9ec0c5ee46d3f61ad661414fb580bf16493358b/config/services.yaml#L20

src/Bundle/DependencyInjection/Compiler/RegisterControllerAliasesPass.php
```diff
@@ -0,0 +1,59 @@
...
+
+use Sylius\Bundle\ResourceBundle\Controller\ResourceController;
+use Sylius\Component\Resource\Metadata\Metadata;
+use Sylius\Component\Resource\Model\ResourceInterface;
+use Symfony\Component\DependencyInjection\Alias;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
+
+final class RegisterControllerAliasesPass implements CompilerPassInterface
```
lchrusciel, line 24: We should adjust the name, as this class is not aliasing controllers anymore
```suggestion
final class RegisterFQCNControllersPass implements CompilerPassInterface
```

lchrusciel:
Hey Loic, 
thanks for the advice. I've totally forgotten about your fix and it was pretty good. However, it works only when the people will customize it in their end application, and from what I've checked `lint:container` command will still complain about resource controllers:
image

In addition, with this fix, you will be able to use the FQCN of controllers in routing:
```yaml
app_book_sortable_index:
    path: /sortable-books/
    methods: [GET]
    defaults:
        _controller: \App\Controller\BookController:indexAction
        _sylius:
            sortable: true
```
@AdamKasp can you add config to your PR? It should be changed in `src/Bundle/test/config/routes.yaml`, line 15th to be exact.

src/Bundle/DependencyInjection/Compiler/RegisterControllerAliasesPass.php
lchrusciel: Lets maybe call it `@internal`, so we can adjust its name in the future
```suggestion
@internal
final class RegisterControllerAliasesPass implements CompilerPassInterface
```

loic425:
> Hey Loic,
> thanks for the advice. I've totally forgotten about your fix and it was pretty good. However, it works only when the people will customize it in their end application, [...]

@lchrusciel Yes, I just wanted you have this little hook on your mind :)

src/Bundle/Tests/DependencyInjection/Compiler/RegisterFQCNControllersPassTest.php
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * @test
+     * @doesNotPerformAssertions
+     */
+    public function it_throws_exception_if_resource_does_not_exist(): void
+    {
+        $this->compile();
+
+        $this->throwException(new InvalidArgumentException());
+    }
```
lchrusciel, line 35: Are you sure about this test?

lchrusciel (approved):

src/Bundle/Tests/DependencyInjection/Compiler/RegisterFQCNControllersPassTest.php
AdamKasp: I hope, that it's testing this part of code:

> try {
            $resources = $container->getParameter('sylius.resources');
        } catch (InvalidArgumentException $exception) {
            return;
        }

Zales0123: What about the check for the second exception that is thrown in the `validateSyliusResource` method?

Zales0123: Maybe we should also check if there is any service registered in the container? I don't know how, to be honest, as we do not set any `sylius.resources` parameter (to check if the exception is thrown) 🤷‍♂️ xD

Zales0123 (approved):

src/Bundle/Tests/DependencyInjection/Compiler/RegisterFQCNControllersPassTest.php
lchrusciel: I didn't dig into it too deeply, but according to https://github.com/sebastianbergmann/phpunit-documentation/issues/171 this is a proper way to do it. Just `throwException` method can be removed.

docs/create_resource.md
```diff
@@ -225,4 +225,6 @@ app_genre_book_add:
                 parameters: { title: resource.title }
 ```
 
+Remember that you can use Fully Qualified Class Name (``\App\Controller\BookController``) instead id ``app.controller.book``
```
pamil, line 228: I guess we don't need the first slash + "of":

```suggestion
Remember that you can use controller's Fully Qualified Class Name (``App\Controller\BookController``) instead of id ``app.controller.book`` 
```

pamil: Same in every doc actually.

src/Bundle/DependencyInjection/Compiler/RegisterFQCNControllersPass.php
```diff
@@ -0,0 +1,59 @@
...
+
+use Sylius\Bundle\ResourceBundle\Controller\ResourceController;
+use Sylius\Component\Resource\Metadata\Metadata;
+use Sylius\Component\Resource\Model\ResourceInterface;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
+
+/** @internal */
+final class RegisterFQCNControllersPass implements CompilerPassInterface
```
pamil, line 24: To comply with PascalCase, this class should be named `RegisterFqcnControllersPass` (same rule as we use for `$resourceId` instead of `$resourceID`).

src/Bundle/DependencyInjection/Compiler/RegisterFQCNControllersPass.php
```diff
@@ -0,0 +1,59 @@
...
+        foreach ($resources as $alias => $configuration) {
+            $metadata = Metadata::fromAliasAndConfiguration($alias, $configuration);
+
+            $this->validateSyliusResource($metadata->getClass('model'));
+            $controllerFQCN = $metadata->getClass('controller');
+
+            if ($controllerFQCN !== ResourceController::class) {
+                $definition = $container->getDefinition($metadata->getServiceId('controller'));
+
+                // TODO: Change to alias definition after bumping to > Symfony 5.2
```
pamil, line 43: Why can't we use aliases in Symfony 4.4?

src/Bundle/DependencyInjection/Compiler/RegisterFQCNControllersPass.php
```diff
@@ -0,0 +1,59 @@
...
+            $resources = $container->getParameter('sylius.resources');
+        } catch (InvalidArgumentException $exception) {
+            return;
+        }
+
+        foreach ($resources as $alias => $configuration) {
+            $metadata = Metadata::fromAliasAndConfiguration($alias, $configuration);
+
+            $this->validateSyliusResource($metadata->getClass('model'));
+            $controllerFQCN = $metadata->getClass('controller');
```
pamil, line 38:
```suggestion
            $controllerFqcn = $metadata->getClass('controller');
```

src/Bundle/Tests/DependencyInjection/Compiler/RegisterFQCNControllersPassTest.php
pamil: I think it doesn't work, since we're extending `AbstractCompilerPassTestCase`, we have this test automatically:

```php
/**
 * This test will run the compile method.
 *
 * @test
 * @coversNothing
 */
final public function compilation_should_not_fail_with_empty_container(): void
{
    try {
        $this->compile();

        // Ensure assertion for strict mode
        $this->assertFalse($this->container->has('no_service'));
    } catch (\Exception $e) {
        $this->fail('The compiler pass should not fail with an empty container.');
    }
}
```

Compiler pass shouldn't throw an exception with an empty container.

pamil: Anyway, what you wanted to do is `$this->expectException()`.

docs/create_resource.md
pamil: Symfony docs say that as well:

```yaml
services:
    # ...

    # explicitly configure the service
    App\Controller\LuckyController:
```

pamil (changes requested):

src/Bundle/DependencyInjection/Compiler/RegisterFQCNControllersPass.php
AdamKasp: Because we need here public aliases which will be provided in Symfony [5.2](https://github.com/symfony/http-kernel/commit/06d999d7a5ade9843bfb613340493ebd752c69d0)

src/Bundle/test/config/routes.yaml
```diff
@@ -12,7 +12,7 @@ app_book_sortable_index:
     path: /sortable-books/
     methods: [GET]
     defaults:
-        _controller: app.controller.book:indexAction
+        _controller: \App\Controller\BookController:indexAction
```
lchrusciel, line 15:
```suggestion
        _controller: App\Controller\BookController:indexAction
```

src/Bundle/Tests/DependencyInjection/Compiler/RegisterFqcnControllersPassTest.php
```diff
@@ -0,0 +1,87 @@
...
+class BookTestClass implements ResourceInterface
+{
+    public function getId()
+    {
+    }
+}
+
+class NormalClass
+{
+}
```
lchrusciel, line 84:
```suggestion
}

```

lchrusciel (approved):

src/Bundle/DependencyInjection/Compiler/RegisterFQCNControllersPass.php
pamil: It's available in Symfony 5.1 as well.

pamil: Could we change the comment? Or maybe put some if to find out if Symfony 5.1 is installed and then using an alias instead of copied definition?

GSadee (approved):

GSadee:
Thanks, Adam! :tada:

PR #328, opened 2021-08-13
Use PHP 7.4 syntax

Zales0123:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Better late than never 💃  I think we should upgrade the PHP syntax whenever we decide to bump the lowest PHP supported version (we would probably be bumping to PHP8.0 in the nearest months, which will also be a good opportunity to improve our codebase 🚀).


lchrusciel (approved):

lchrusciel:
Thank you, Mateusz! :1st_place_medal:

PR #342, opened 2021-12-20
Fix CI

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT


phpstan.neon
```diff
@@ -9,8 +9,8 @@ parameters:
     checkMissingIterableValueType: false
     reportUnmatchedIgnoredErrors: false
 
-    excludes_analyse:
-        - %currentWorkingDirectory%/src/Bundle/Controller/ControllerTrait.php
+    excludePaths:
+        - %currentWorkingDirectory%/src/Bundle/Controller/*
```
Zales0123, line 13: 👍 for now, but it would be good to come back to this change after the merge and figure out how to exclude as few classes as possible

src/Bundle/Controller/Parameters.php
```diff
@@ -14,23 +14,61 @@
...
-class Parameters extends ParameterBag
-{
-    /**
-     * @param string $key
-     * @param mixed $default
-     *
-     * @return mixed
-     */
-    public function get($key, $default = null)
+if (Kernel::MAJOR_VERSION === 6) {
```
Zales0123, line 19: I believe we can skip check for Symfony 6 for now, fix the build with Symfony 4.4 and 5.4 **and then** drop 4.4 and introduce Symfony 6 properly, wdyt?

src/Bundle/Controller/Parameters.php
```diff
@@ -14,23 +14,61 @@
...
+            $result = parent::get($key, $default);
+
+            if (null === $result && $default !== null && $this->has($key)) {
+                $result = $default;
+            }
+
+            return $result;
         }
+    }
+
```
Zales0123, line 53: Extra blank line 😄

Zales0123 (commented):
👌

src/Bundle/Controller/Parameters.php
loic425: You're right... it's done.

src/Bundle/Controller/Parameters.php
loic425: fixed

loic425:
@Zales0123 It's green now :)

Zales0123:
Thank you, @loic425! 🎉

PR #355, opened 2022-01-17
Tests for index and show in UI

Zales0123:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

This bundle is a very important fundamental for Sylius itself, but at the same time, it provides so many untested functionalities... It's just a shame 😄 I propose two simple UI-based tests for the resource. To speed up implementation, I've used `ApiTestCase` to load fixtures quickly, but we should think about extracting this fixtures logic (probably in the `api-test-case` itself) and use some crawler for these tests. Let me know what you think about it  🖖


src/Bundle/test/src/Entity/ScienceBook.php
```diff
@@ -0,0 +1,60 @@
...
+
+declare(strict_types=1);
+
+namespace App\Entity;
+
+use Sylius\Component\Resource\Model\ResourceInterface;
+
+class ScienceBook implements ResourceInterface
+{
+    private ?int $id;
```
loic425, line 20:
```suggestion
    private ?int $id = null;
```

src/Bundle/test/src/Entity/ScienceBook.php
```diff
@@ -0,0 +1,60 @@
...
+
+    private ?Author $author = null;
+
+    private ?string $title = null;
+
+    public function getId()
```
loic425, line 26:
```suggestion
    public function getId(): ?int
```

src/Bundle/test/src/Entity/ScienceBook.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        return $this->id;
+    }
+
+    public function getTitle(): ?string
+    {
+        return $this->title;
+    }
+
+    public function setTitle($title): void
```
loic425, line 36:
```suggestion
    public function setTitle(?string $title): void
```

src/Bundle/test/src/Entity/ScienceBook.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        $this->title = $title;
+    }
+
+    public function getAuthorFirstName(): ?string
+    {
+        return $this->author ? $this->author->getFirstName() : null;
```
loic425, line 43:
```suggestion
        return $this->author?->getFirstName();
```

src/Bundle/test/src/Entity/ScienceBook.php
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public function getAuthorLastName(): ?string
+    {
+        return $this->author ? $this->author->getLastName() : null;
```
loic425, line 48:
```suggestion
        return $this->author?->getLastName();
```

loic425 (changes requested):

src/Bundle/test/src/Entity/ScienceBook.php
vvasiloi: I have been waiting for this syntax for so long...
It's annoying especially when using Doctrine and everything is nullable.

loic425 (approved):

PR #359, opened 2022-01-20
Fix type of serialization groups

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | 
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Component/Annotation/SyliusRoute.php
+++ b/src/Component/Annotation/SyliusRoute.php
@@ -30,7 +30,7 @@ final class SyliusRoute
 
     public ?array $criteria = null;
 
-    public ?string $serializationGroups = null;
+    public ?array $serializationGroups = null;
 
     public ?string $serializationVersion = null;
 
@@ -54,7 +54,7 @@ final class SyliusRoute
         ?string $template = null,
         ?array $repository = null,
         ?array $criteria = null,
-        ?string $serializationGroups = null,
+        ?array $serializationGroups = null,
         ?string $serializationVersion = null,
         ?array $requirements = null,
         ?array $options = null,


Roshyo (approved):

vvasiloi (approved):

lchrusciel (approved):
Can we write super simple case for that to avoid issues in the future?

loic425:
> Can we write super simple case for that to avoid issues in the future?

There's already a simple test.... But I don't even know how it keeps working with the wrong type.
https://github.com/Sylius/SyliusResourceBundle/blob/master/src/Bundle/Tests/Routing/RoutesAttributesLoaderTest.php#L147

lchrusciel:
thats the worst type of bugs - you know, that tests should fail, but they are green :(

loic425:
> thats the worst type of bugs - you know, that tests should fail, but they are green :(

Yes but the result is ok, even with the wrong type, I also test it manually 🗡️

loic425:
@lchrusciel I think there is some "magic" somewhere :)

vvasiloi:
It works... must be a bug (or feature) of PHP: https://3v4l.org/fOejp#v8.0.15

<details>
<summary>Screenshot</summary>
image
</details>

vvasiloi:
And although I only defined the `serializationGroups` in the attribute class, I also passed (unintentionally, copy/paste) name, path and controller and all were returned as a result.

lchrusciel (approved):

lchrusciel:
Thanks, Loïc! :1st_place_medal:

PR #363, opened 2022-02-01
Fix docs for templates customization

### LINKED ISSUE #362: Incorrect documentation for overriding the view of a resource form
latest

The [config reference](https://github.com/Sylius/SyliusResourceBundle/blob/master/docs/reference.md) suggests to overwrite a resource's form rendering as such
```
sylius_resource:
    resources:
        app.book:
            driver: doctrine/orm
            classes:
            options:
                object_manager: default
            templates:
                form: Book/_form.html.twig
```
However [Configuration.php](https://github.com/Sylius/SyliusResourceBundle/blob/master/src/Bundle/DependencyInjection/Configuration.php) suggests the only acceptable value is a scalar
```
 private function addResourcesSection(ArrayNodeDefinition $node): void
    {
        $node
            ->children()
                ->arrayNode('resources')
                    ->useAttributeAsKey('name')
                    ->arrayPrototype()
                        ->children()
                            ->scalarNode('driver')->defaultValue(SyliusResourceBundle::DRIVER_DOCTRINE_ORM)->end()
                            ->variableNode('options')->end()
                            ->scalarNode('templates')->cannotBeEmpty()->end()
```
So when I configure:
```
sylius_resource:
  resources:
etc etc etc
      templates:
        form: Admin/Ruleset/_form.html.twig
```
I get:
`Invalid type for path "sylius_resource.resources.skyboundtech.wholesale_ruleset.templates". Expected "scalar", but got "array".`

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no 
| Related tickets | fixes #362 
| License         | MIT

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -15,8 +15,7 @@ sylius_resource:
                     validation_groups: [sylius]
             options:
                 object_manager: default
-            templates:
-                form: Book/_form.html.twig
+            templates: 'book' # Templates namespace for this resource, default value is empty
             translation:
                 classes:
                     model: ~
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -15,7 +15,7 @@ sylius_resource:
                     validation_groups: [sylius]
             options:
                 object_manager: default
-            templates: 'book' # Templates namespace for this resource, default value is empty
+            templates: 'book' # directory of templates, this would be placed in `templates/book/`. Default value is empty. Keep in mind, that in Sylius we usually use UpperCamelCase notation in contrary to Symfony recommendations
             translation:
                 classes:
                     model: ~
--- a/src/Bundle/Tests/Configuration/ConfigurationTest.php
+++ b/src/Bundle/Tests/Configuration/ConfigurationTest.php
@@ -13,9 +13,13 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\ResourceBundle\Tests\Configuration;
 
+use App\Entity\Book;
 use Matthias\SymfonyConfigTest\PhpUnit\ConfigurationTestCaseTrait;
 use PHPUnit\Framework\TestCase;
+use Sylius\Bundle\ResourceBundle\Controller\ResourceController;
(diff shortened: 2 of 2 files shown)


docs/reference.md
```diff
@@ -15,8 +15,7 @@ sylius_resource:
                     validation_groups: [sylius]
             options:
                 object_manager: default
-            templates:
-                form: Book/_form.html.twig
+            templates: 'book' # Templates namespace for this resource, default value is empty
```
Roshyo, line 18: We should perhaps be more specific about what's expected here ? Like `Book/Crud/` 
WDYT ?

Roshyo: I mean, just this `book` may be too simple for the users to understand what's expected

Zales0123: 👍 for `Book/Crud/`. Additionally, I would maybe change the comment to `Templates **directory** (...)`? As templates do not have namespace _per se_, as e.g. php classes 😄 We can even go wth some example, like:

```yaml
templates: 'Book/Crud/' # directory of templates, this would be placed in `templates/Book/Crud/`. Default value is empty
```

or something similar, wdyt?

loic425: @Zales0123 I'm ok, but using snake case, this is the Symfony best practice recommandation.
`templates: 'book/crud/' # directory of templates, this would be placed in `templates/book/crud/`. Default value is empty`

Roshyo: I agree with that, but it is not what Sylius does everywhere for now.
Perhaps it is time to initiate that change :smile:

loic425: For Sylius, I don't know, but this package is not only used on Sylius projects, so, I think we should use the recommandation.

lchrusciel: perhaps we can add two recommendations? 
```suggestion
            templates: 'book' # directory of templates, this would be placed in `templates/book/`. Default value is empty. Keep in mind, that in Sylius we usually use UpperCamelCase notation in contrary to Symfony recommendations
```

PR #364, opened 2022-02-02
Remove resource when redirect on delete

Jibbarth:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Hi team :wave: 

I add this issue when trying to call deleteAction on a custom entity. 

The fact is my entity is strongly typed, and there is no nullable id on it : 

```php
class Entity implement ResourceInterface
{ 
    private int $id;
    
    public function getId(): int
    { 
        return $this->id;
    }
}
```

Passing the resource on this redirectToIndex force it to look in the resource for any id parameters ([here](https://github.com/Sylius/SyliusResourceBundle/blob/master/src/Bundle/Controller/RequestConfiguration.php#L535)), and it's raised an exception while trying to access to a property not initialized.

I guess on a deleteAction, it's not mandatory to retrieve the previous id, so we can safely remove it ? 

I just submit quickly this patch to see if every tests passes, if it's not the case, i'll look deeper what happen. 
WDYT ?

### DIFF
--- a/src/Bundle/Controller/ResourceController.php
+++ b/src/Bundle/Controller/ResourceController.php
@@ -374,7 +374,7 @@ class ResourceController
             return $postEventResponse;
         }
 
-        return $this->redirectHandler->redirectToIndex($configuration, $resource);
+        return $this->redirectHandler->redirectToIndex($configuration);
     }
 
     public function bulkDeleteAction(Request $request): Response
--- a/src/Bundle/spec/Controller/ResourceControllerSpec.php
+++ b/src/Bundle/spec/Controller/ResourceControllerSpec.php
@@ -1923,7 +1923,7 @@ final class ResourceControllerSpec extends ObjectBehavior
         $postEvent->getResponse()->willReturn(null);
 
         $flashHelper->addSuccessFlash($configuration, ResourceActions::DELETE, $resource)->shouldBeCalled();
-        $redirectHandler->redirectToIndex($configuration, $resource)->willReturn($redirectResponse);
+        $redirectHandler->redirectToIndex($configuration)->willReturn($redirectResponse);
 
         $this->deleteAction($request)->shouldReturn($redirectResponse);
     }


Jibbarth:
:thinking: It went bad ! I'll check why phpspec is not happy :sweat_smile:

vvasiloi:
@loic425 WDYT about this one?

Roshyo (approved):

vvasiloi:
> I guess on a deleteAction, it's not mandatory to retrieve the previous id, so we can safely remove it ?
It is optional, but I don't think it should be removed.
A valid use case for it is having configured a custom redirect with parameters values supplied from the resource.
I'm not even sure there is an issue... after all, the ID is supposed to be nullable by design.

Zales0123:
Hello @Jibbarth, sorry for the long response time 🖖 Additional to the @vvasiloi comment I would add, that in spite of is it needed or not, we need also to think about backward compatibility... which means, it **can already be used** by someone using Resource Bundle :/ So removing it, if it's not very important to fix a critical bug, should be avoided 🌧️ 

Nevertheless, we need to decide, whether we want to remove it (and then document a potential BC break) or keep it and require nullable id on the resource entity (and then we need to document it properly). cc @lchrusciel @loic425

lchrusciel:
Hey Jibe,

thanks for your proposal. I can only bump @vvasiloi response. I'm not sure, how you create your entities, but if it is handled with DoctrineORM, then ID by design is nullable. Doctrine will remove it with reflection during deletion. I know it is annoying, but this is how it works.

However, should we look for "id" after deletion? Probably not. So I would not remove passing a resource into a redirect, rather change the default behaviour to not request an ID

vvasiloi:
This behavior comes from these lines:
https://github.com/Sylius/SyliusResourceBundle/blob/9192ee9f1248ee9a9855788763330c1649713d62/src/Bundle/Controller/RequestConfiguration.php#L534-L536
In my opinion, that gives the `parseResourceValues` an extra responsibility it should not have.
This logic should be moved outside of this function, probably to its only caller `getRedirectParameters`.
Unfortunately, there's no way to determine here if the route actually needs the id parameter.
That means we have to go one more level up, to the `RedirectHandler`.
Here it makes sense to add the parameter to `redirectToResource` and keep it out of `redirectToIndex`.
Even better, we have the router here, so we can check if the route actually needs the id parameter.

The whole thing was probably meant to ease developers' lives, so they don't have to explicitly pass the id parameter every time.
The downside is a bit messy code that also adds unexpected `?id=123` parameters in customer-facing redirect URLs.

lchrusciel:
this default behaviour is quite useful for create/update redirects to update/show pages. This behaviour would need to be left untouched

vvasiloi:
> this default behaviour is quite useful for create/update redirects to update/show pages

I suspect it was specifically added for the resource route loader, where you can configure it like `redirect: update`.
If that's the case, then it can be handled in the route loader by adding the parameter explicitly, in the form of `['id' => 'resource.id']`.
I have to look at the history and track down the change maybe there's more context.
In any way, I'm pretty sure we can make it better.

Jibbarth:
Closing as it could be BC. Thanks all

Issue #335, opened 2021-11-12
Symfony deprecations


dannyvw:
Since symfony/dependency-injection 5.1: The signature of method "Symfony\Component\DependencyInjection\Alias::setDeprecated()" requires 3 arguments: "string $package, string $version, string $message", not defining them is deprecated.

`->setDeprecated(true, 'The "%alias_id%" service alias is deprecated since Sylius 1.8, use the "pagerfanta.twig_extension" service ID instead.');`

dannyvw:
This deprecation still exist

User Deprecated: Method "Symfony\Component\Form\DataTransformerInterface::transform()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "Sylius\Bundle\ResourceBundle\Form\DataTransformer\ResourceToIdentifierTransformer" now to avoid errors or add an explicit @return annotation to suppress this message.

PR #379, opened 2022-03-12
Fix setDeprecated deprecation

dannyvw:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes 
| Related tickets | partially #335 
| License         | MIT

### DIFF
--- a/src/Bundle/DependencyInjection/Compiler/PagerfantaBridgePass.php
+++ b/src/Bundle/DependencyInjection/Compiler/PagerfantaBridgePass.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler;
 
+use Symfony\Component\DependencyInjection\Alias;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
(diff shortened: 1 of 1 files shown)


vvasiloi (approved):

src/Bundle/DependencyInjection/Compiler/PagerfantaBridgePass.php
```diff
@@ -48,12 +49,23 @@ private function aliasRenamedServices(ContainerBuilder $container): void
...
-                ->setDeprecated(true, 'The "%alias_id%" service alias is deprecated since Sylius 1.8, use the "pagerfanta.view_factory" service ID instead.');
+                ->setDeprecated(...$this->getDeprecationMessage());
         }
     }
+
+    private function getDeprecationMessage(): array
+    {
+        $message = 'The "%alias_id%" service alias is deprecated since Sylius 1.8, use the "pagerfanta.view_factory" service ID instead.';
+
+        if (method_exists(Alias::class, 'getDeprecation')) {
```
lchrusciel, line 65: Won't it be better to base it on Symfony version?

dannyvw: The `getDeprecation` method is added in the same PR, see https://github.com/symfony/symfony/pull/35778. So i think this should be good enough.

lchrusciel: I get it, I have the problem, that checking the number of arguments or Symfony version is a more explicit check for me. But perhaps, I'm just picky

lchrusciel:
Thank you, Danny! :1st_place_medal:

PR #380, opened 2022-03-13
Fix session deprecation

dannyvw:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes 
| Related tickets | partially #335
| License         | MIT

### DIFF
--- a/src/Bundle/Controller/FlashHelper.php
+++ b/src/Bundle/Controller/FlashHelper.php
@@ -16,6 +16,7 @@ namespace Sylius\Bundle\ResourceBundle\Controller;
 use Sylius\Bundle\ResourceBundle\Event\ResourceControllerEvent;
 use Sylius\Component\Resource\Metadata\MetadataInterface;
 use Sylius\Component\Resource\Model\ResourceInterface;
+use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Session\Flash\FlashBagInterface;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\Translation\TranslatorBagInterface;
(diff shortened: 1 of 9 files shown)


Roshyo (approved):

src/Bundle/DependencyInjection/Compiler/SessionPass.php
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\HttpFoundation\RequestStack;
+
+/**
+ * TODO Remove after bumping to > Symfony 5.x
```
lchrusciel, line 22: What about bumping minimal requirements to 4.4 already?

lchrusciel:
Thank you, Danny! :tada:

src/Bundle/Controller/FlashHelper.php
```diff
@@ -89,8 +103,14 @@ private function addFlash(string $type, string $message, array $parameters = [])
             $message = $this->prepareMessage($message, $parameters);
         }
 
+        if ($this->requestStack instanceof SessionInterface) {
```
lchrusciel, line 106: Having afterthought, perhaps we should wrap it into our custom class, that would do this logic?

src/Bundle/DependencyInjection/Compiler/SessionPass.php
Roshyo: If you meant 5.4, then I agree

PR #404, opened 2022-04-04
Add support for DTOs

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

DTO support for create and update operations.

## Routes
```yaml
app_register_user:
    path: /users/register
    methods: [POST]
    defaults:
        _controller: app.controller.user:createAction
        _sylius:
            input: App\Dto\Register
            output: App\Dto\User
            form: App\Form\Type\RegisterType
```

```php
<?php

namespace App\Entity;

use App\Dto\Register;
use App\Dto\User as DtoUser;

#[SyliusRoute(
    name: 'register',
    path: '/register',
    methods: ['GET', 'POST'],
    controller: 'app.controller.user:createAction',
    form: RegisterType::class,
    input: Register::class,
    output: DtoUser::class
)]
class User
```

## Data transformers
```php
<?php

declare(strict_types=1);

namespace App\DataTransformer;

use App\Dto\Register;
use App\Entity\User;
use Sylius\Bundle\ResourceBundle\Controller\RequestConfiguration;
use Sylius\Bundle\ResourceBundle\DataTransformer\DataTransformerInterface;
use Webmozart\Assert\Assert;

final class RegisterToUserEntityDataTransformer implements DataTransformerInterface
{
    public function transform(object $data, string $to, RequestConfiguration $configuration): User
    {
        Assert::isInstanceOf($data, Register::class);

        $user = new User();
        $user->setUsername($data->username);
        $user->setPassword($data->password);

        return $user;
    }

    public function supportsTransformation(object $data, string $to, RequestConfiguration $configuration): bool
    {
        return $data instanceof Register && is_a($to, User::class, true);
    }
}

```

```php
<?php

declare(strict_types=1);

namespace App\DataTransformer;

use App\Dto\User;
use App\Entity\User as UserEntity;
use Sylius\Bundle\ResourceBundle\Controller\RequestConfiguration;
use Sylius\Bundle\ResourceBundle\DataTransformer\DataTransformerInterface;
use Webmozart\Assert\Assert;

final class UserEntityToDtoDataTransformer implements DataTransformerInterface
{
    public function transform(object $data, string $to, RequestConfiguration $configuration): User
    {
        Assert::isInstanceOf($data, UserEntity::class);

        return new User($data->getUsername() ?: '');
    }

    public function supportsTransformation(object $data, string $to, RequestConfiguration $configuration): bool
    {
        return $data instanceof UserEntity && is_a($to, User::class, true);
    }
}

```

* [x] create
    * [x] input
    * [x] output 
* [x] update
    * [x] input
    * [x] output 
    
* [x] Remove bc-break with resource form factory


src/Bundle/test/src/Dto/UpdateProfile.php
```diff
@@ -0,0 +1,24 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace App\Dto;
+
+final class UpdateProfile
+{
+    public ?string $username = null;
```
lchrusciel, line 18: can't we use constructor promotion in these tests?

src/Bundle/DependencyInjection/Compiler/ChainDataTransformerPass.php
```diff
@@ -0,0 +1,33 @@
...
+final class ChainDataTransformerPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        $definition = $container->findDefinition(ChainDataTransformerInterface::class);
+        $taggedServices = $container->findTaggedServiceIds('sylius.data_transformer');
+
+        foreach ($taggedServices as $id => $tags) {
+            // add the data transformer service to the ChainDataTransformer service
+            $definition->addMethodCall('addDataTransformer', [new Reference($id)]);
```
lchrusciel, line 30: instead of the usage of a method call, can we define `ChainDataTransformerInterface` as a regular service and inject in to It tagged_iterator? I would prefer to avoid method injection if we can

src/Bundle/DataTransformer/ChainDataTransformer.php
```diff
@@ -0,0 +1,57 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ResourceBundle\DataTransformer;
+
+use Sylius\Bundle\ResourceBundle\Controller\RequestConfiguration;
+
+final class ChainDataTransformer implements ChainDataTransformerInterface
```
lchrusciel, line 18: That's a neat solution, I like it!. It is heavily inspired by ApiPlatform, isn't it? 

One thing, I've learned from Sylius integration is to keep the possibility to have more than one data transformer for objects. Thanks to that, you can add some common metadata to objects/commands like data from the configuration, attributes or logged in user without calling it from the inside the application. I called it the "CommandEnrichment" in Sylius and perhaps we could implement it right away. The main idea is to keep running datatransformers after the first strike. On the other hand, we may make distinct concept for it

loic425: @lchrusciel Yes, all this PR is inspired by ApiPlatform implementation 🥇 
There's no ChainDataTransformer on ApiPlatform, but the data transformer system is the same as the ApiPlatform one.

src/Bundle/Controller/ResourceController.php
```diff
@@ -202,6 +234,10 @@ public function createAction(Request $request): Response
 
             $postEvent = $this->eventDispatcher->dispatchPostEvent(ResourceActions::CREATE, $configuration, $newResource);
 
+            if (null !== $output = $configuration->getOutput()) {
```
loic425, line 237: @lchrusciel  Here, I do not use the resource factory (to create the Doctrine entity) and the data transformer (to transform into the DTO), cause we can directly use a null value on the form.

I'm not sure if we have to add a way to use a factory method for the DTO or sth else.

@pocky WDTY ?

src/Bundle/Controller/ResourceController.php
```diff
@@ -240,7 +276,11 @@ public function updateAction(Request $request): Response
         $this->isGrantedOr403($configuration, ResourceActions::UPDATE);
         $resource = $this->findOr404($configuration);
 
-        $form = $this->resourceFormFactory->create($configuration, $resource);
+        if (null !== $input = $configuration->getInput()) {
+            $resource = $this->transform($resource, $input, $configuration);
```
loic425, line 280: I think the `assertNotNull' is missing

src/Bundle/test/src/Dto/UpdateProfile.php
loic425: done

src/Bundle/DependencyInjection/Compiler/ChainDataTransformerPass.php
loic425: done

src/Bundle/Controller/ResourceController.php
loic425: Hum this is the same as `createAction`
to update the password, for example, we do not want to instanciate any transformation and pass null data to the form.

loic425:
@lchrusciel @Zales0123 I listened to the conf for the new system in API platform (2.7 & 3.0) on Twitter which explain the
data transformer deprecation. So I think this is both a better and simple way for the resource controller too, so I'll make another PR. I read the new documentation and this exactly where I want to go with messenger. Let's make an awesome release with these features 🐰 🏎️

loic425:
I've just created an ADR, to expose the idea for State Providers and Processors.
#456

loic425:
We have the new resource metadata architecture now.

PR #381, opened 2022-03-16
Bump Pagerfanta from 2.x to 3.x

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

PR #415, opened 2022-04-08
Fix pagerfanta requirements

gseidel:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Since version `1.8.3` the requirement `babdev/pagerfanta-bundle` was upgraded to version `^3.0`. But this version requires `pagerfanta/core` instead of `pagerfanta/pagerfanta`. So in the SyliusResourceBundle we have missing classes like `Pagerfanta\Adapter\DoctrineORMAdapter` which is not part of the pagerfanta core. This leads to no more working code if people don't install `pagerfanta/pagerfanta` their own. This PR will fix this requirements.


mbabker:
If you’re going to add anything to the requirements, it really should just be the `pagerfanta/doctrine-orm-adapter` package and not the `pagerfanta/pagerfanta` monorepo (using the split packages installs a lot less code and has better dependency constraints than using the monorepo can ever have).

gseidel:
Thanks [mbabker](https://github.com/mbabker). I was not aware of the pagerfanta packages. To just require the needed package make a lot more sense. PR is updated

gseidel:
After checking the code I found other dependencies requiring speciefied subpackage of pagerfanta
- `pagerfanta/doctrine-mongodb-odm-adapter`
- `pagerfanta/doctrine-phpcr-odm-adapter`
- `pagerfanta/twig`

For classes:
- `Pagerfanta\Doctrine\MongoDBODM\QueryAdapter`
- `Pagerfanta\Doctrine\PHPCRODM\QueryAdapter`
- `Pagerfanta\Twig\Extension\PagerfantaExtension`

The main problem is that the adapter classes will also install further packages, which `pagerfanta/pagerfanta` doesn't

I would suggest to revert my commit and use `pagerfanta/pagerfanta` for the branch 1.8 and in other versions where we could make a BC, we can use specified pagerfanta packages.

I can remeber that the resource-bundle will drop Mongo and CR support in higher versions. So we can easily add the orm package then. Otherwise a user have to install a pagerfanta package himself, when he will install the resource-bundle

loic425:
I think to avoid bc-break we can move pagerfanta/pagerfanta on main requirements on 1.8 & 1.9.
we can handle the "optional packages" on 1.10
WDTY @Zales0123?

mbabker:
The problem here is no different than Symfony itself with its optional dependencies.  Actually, the Pagerfanta and Symfony monorepos are kind of the same in that they don’t install all of the optional dependencies and users are still forced to make sure the full dependency stacks are in place (the Pagerfanta monorepo doesn’t install any of the Doctrine packages or Twig, but the subsplits do have those dependencies declared).

Using the monorepo is fine, but from a downstream consumer perspective, it will cause issues by not installing all required transient dependencies and it won’t adequately lock version ranges (you can install the ORM 3.0 branch with the monorepo and it won’t raise a flag, but do the same with the ORM adapter package and you’ll get Composer warnings).

Short term, using the monorepo might “fix” any immediate issues, but long term folks really should stop using it.

mbabker:
Hmm, I didn't realize another 1.8 release was made alongside 1.9 and that the Pagerfanta major version bump was included in that.  #381 IMO should be reverted from 1.8, having a major version bump on a dependency in a patch release seems like it's asking for trouble.  If anything, https://github.com/Sylius/SyliusResourceBundle/commit/a51e1a26489fc900d44c62c092ebce582857b9d7 (which was the initial version of #298 before it was changed to 3.0 only) should be used which allows both Pagerfanta 2.x and 3.x to be installed.

src/Bundle/Tests/DependencyInjection/Compiler/PagerfantaBridgePassTest.php
```diff
@@ -26,7 +25,7 @@ final class PagerfantaBridgePassTest extends AbstractCompilerPassTestCase
      */
     public function it_creates_aliased_services_and_changes_the_view_factory_class(): void
     {
-        $this->registerService('pagerfanta.twig_extension', PagerfantaExtension::class);
+        $this->registerService('pagerfanta.twig_extension', class_exists('Pagerfanta\Twig\Extension\PagerfantaExtension') ? 'Pagerfanta\Twig\Extension\PagerfantaExtension' : 'BabDev\PagerfantaBundle\Twig\PagerfantaExtension');
```
mbabker, line 28: You can always use `Pagerfanta\Twig\Extension\PagerfantaExtension` here.  Since the bundle's 2.5 release, that's the class used for the `pagerfanta.twig_extension` service, and with the version constraints given, it'll always be available.

mbabker (approved):

lchrusciel:
Hey folks,

thanks a lot for your contribution @gseidel. Your first in this repository! I wish it would be in better circumstances ;) 

But coming back to the implementation itself, I fail to see how this change would help you fix your repository. According to the change, a composer will try to solve the highest possible dependencies, so you still will have to define the proper pagerfanta version, or am I wrong? 

it seems, that the only revert proposed by @mbabker will solve 1.8 compatibility issues. Next question is, how broke is the 1.9 release.

gseidel:
Hello @lchrusciel. Don't worry about the circumstances, I using this bundle for a long time already and finally I can contribute ;)

This PR will fix two cases. My case is, that I can upgrade to `pagerfanta-bundle` version `3.*` but if I upgrade a bug fix version I would not except to manully add a package. So I moved the `pagerfanta/pagerfanta` from dev to requirement.

On the other hand, their may be people who can't upgrade to version `3.*`, because they have an other package which stick to version `2.*` and the bump version will fail for them, because they can't resolve the dependency graph. So I follow @mbabker suggestion and allow version `2.*` as well.

lchrusciel:
@gseidel can you check your app with https://github.com/Sylius/SyliusResourceBundle/releases/tag/v1.8.4?

gseidel:
I close this PR and rebase it to branch `1.9`. See #420

PR #416, opened 2022-04-11
Revert "Bump Pagerfanta from 2.x to 3.x"

lchrusciel:
Reverts Sylius/SyliusResourceBundle#381

Alternative solution for #415

PR #417, opened 2022-04-11
Revert "Revert "Bump Pagerfanta from 2.x to 3.x""

lchrusciel:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | reverts #416 for higher branches
| License         | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,6 +58,7 @@
         "doctrine/orm": "^2.5",
         "lchrusciel/api-test-case": "^5.0",
         "matthiasnoback/symfony-dependency-injection-test": "^4.2.1",
+        "pagerfanta/pagerfanta": "^3.0",
         "pamil/phpspec-skip-example-extension": "^4.2",
         "phpspec/phpspec": "^7.2",
         "phpspec/prophecy-phpunit": "^2.0",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -58,6 +58,7 @@
         "doctrine/orm": "^2.5",
         "lchrusciel/api-test-case": "^5.0",
         "matthiasnoback/symfony-dependency-injection-test": "^4.2.1",
+        "pagerfanta/pagerfanta": "^3.0",
```
loic425, line 61: for now I suggest you to move it on main requirements. We'll handle the "optional" requirements later on next release

loic425 (changes requested):

GSadee (approved):

composer.json
lchrusciel: it was just a revert to make our branches green again, so I would go with it for now. But I agree, that 1.9 needs this. The question is should be part of this PR?

loic425: Hum good idea to make this change in another "Fix" PR.

loic425 (approved):

lchrusciel:
Thank you, Łukasz! :tada:

PR #418, opened 2022-04-11
[Symfony 6] Fix getting container on PHPUnit tests

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes (for Symfony 6)
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Bundle/Tests/Routing/CrudRoutesAttributesLoaderTest.php
+++ b/src/Bundle/Tests/Routing/CrudRoutesAttributesLoaderTest.php
@@ -24,7 +24,7 @@ final class CrudRoutesAttributesLoaderTest extends KernelTestCase
     {
         self::bootKernel(['environment' => 'test_with_attributes']);
 
-        $container = static::$container;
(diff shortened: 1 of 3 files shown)


src/Bundle/Tests/Routing/CrudRoutesAttributesLoaderTest.php
```diff
@@ -24,7 +24,7 @@ public function it_generates_crud_routes_from_resource_alias(): void
     {
         self::bootKernel(['environment' => 'test_with_attributes']);
 
-        $container = static::$container;
+        $container = \method_exists($this, 'getContainer') ? static::getContainer() : static::$container;
```
lchrusciel: for which version do we need this check? Since the drop of Sf4.4 we should be fine with just refactoring to "getContainer", won't we?

loic425: done

lchrusciel (approved):

PR #423, opened 2022-04-12
Make resource transformer

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #404 
| License         | MIT


src/Bundle/Command/StubMakeResourceTransformer.php
```diff
@@ -0,0 +1,19 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ResourceBundle\Command;
+
+final class StubMakeResourceTransformer extends StubCommand
+{
+    protected static $defaultName = 'make:resource-transformer';
```
lchrusciel, line 18: before that I would even introduce make:resource command :)

src/Bundle/Command/StubCommand.php
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ResourceBundle\Command;
```
lchrusciel, line 14: Perhaps we should adjust namespace to something like CLI to avoid name collision with command/events from CQRS?
```suggestion
namespace Sylius\Bundle\ResourceBundle\CLI;
```

loic425: I am opiniated cause `Command` is the Symfony naming convention for console commands.

src/Bundle/Command/StubMakeResourceTransformer.php
loic425: Yes, one day... 🤣

loic425:
This is not needed anymore, but we'll need some other makers.

Issue #400, opened 2022-03-29
Symfony 6 support


loic425:
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/411
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/412
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/401
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/413
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/398
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/418
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/399
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/419
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/421
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/422
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/437
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/436
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/438
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/439
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/440
* [x] https://github.com/Sylius/SyliusResourceBundle/pull/444

And the pull request to rule them all https://github.com/Sylius/SyliusResourceBundle/pull/428

PR #444, opened 2022-04-20
 [Symfony 6] Third solution to fix resource controllers

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes (for Symfony 6)
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | partially #400
| License         | MIT

### DIFF
--- a/src/Bundle/DependencyInjection/Compiler/TwigPass.php
+++ b/src/Bundle/DependencyInjection/Compiler/TwigPass.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
 declare(strict_types=1);
 
 namespace Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler;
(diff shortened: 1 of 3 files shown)


src/Bundle/DependencyInjection/Compiler/TwigPass.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+final class TwigPass implements CompilerPassInterface
```
lchrusciel, line 22: Let's add note, that this class will be removed in Resource 2.0

lchrusciel (approved):

src/Bundle/DependencyInjection/Compiler/TwigPass.php
loic425: done

GSadee: Shouldn't we deprecate this class and inform about removing setting twig as public service in the UPGRADE file?

loic425: I think `deprecate it` should be landed with Sylius resource controller split (next release). Cause we need this usage for this release and can not remove using it.

lchrusciel:

PR #450, opened 2022-05-04
Adjust when some runtime deprecation notices are triggered and use Symfony's `trigger_deprecation()` helper

### LINKED ISSUE #448: Change deprecations
Can we change some deprecations? Deprecations like this are rendered on each page but we can't solve them

https://github.com/Sylius/SyliusResourceBundle/blob/1.10/src/Bundle/DependencyInjection/PagerfantaExtension.php#L20
https://github.com/Sylius/SyliusResourceBundle/blob/1.10/src/Bundle/DependencyInjection/PagerfantaConfiguration.php#L20
https://github.com/Sylius/SyliusResourceBundle/blob/1.10/src/Bundle/DependencyInjection/Compiler/PagerfantaBridgePass.php#L20
https://github.com/Sylius/SyliusResourceBundle/blob/1.10/src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineODMDriver.php#L25
https://github.com/Sylius/SyliusResourceBundle/blob/1.10/src/Bundle/DependencyInjection/Driver/Doctrine/DoctrinePHPCRDriver.php#L27

mbabker:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #448
| License         | MIT

This PR will do two things:

1) Adjusts when some runtime deprecations (especially for those which are unfixable in consuming applications) to be triggered when the deprecated element is actually used instead of arbitrarily emitting them when the file is loaded
2) Uses Symfony's `trigger_deprecation()` helper for runtime deprecations; this is designed to be a "pluggable" function that can allow someone to customize the functionality in their applications if so desired (plus the function's signature forces key information at the front of each message which makes scanning the messages a bit better IMO)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,6 +37,7 @@
         "jms/serializer-bundle": "^3.5 || ^4.0",
         "sylius/registry": "^1.2",
         "symfony/config": "^5.4 || ^6.0",
+        "symfony/deprecation-contracts": "^2.1 || ^3.0",
         "symfony/expression-language": "^5.4 || ^6.0",
         "symfony/form": "^5.4 || ^6.0",
         "symfony/framework-bundle": "^5.4 || ^6.0",
--- a/src/Bundle/AbstractResourceBundle.php
+++ b/src/Bundle/AbstractResourceBundle.php
@@ -104,10 +104,12 @@ abstract class AbstractResourceBundle extends Bundle implements ResourceBundleIn
     {
         switch ($driverType) {
             case SyliusResourceBundle::DRIVER_DOCTRINE_MONGODB_ODM:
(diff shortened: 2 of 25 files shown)


loic425 (approved):

loic425:
@lchrusciel I think it should be based on bug fix branch (1.9)

Zales0123 (approved):
Seems as a good DX improvement 👍

lchrusciel (approved):

Zales0123:
Thank you, @mbabker! 🏅

PR #459, opened 2022-06-23
Fix coding standards to fix the build

Zales0123:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

EDIT: I had to upgrade PHPStan to a more modern version (we should've done it a long time ago :dancer:) to make it work with PHP 8.1. Therefore, it resulted in a little bit more changes and skips than expected, but at least it unblocks us with the development 🖖

### DIFF
--- a/src/Bundle/Event/ResourceControllerEvent.php
+++ b/src/Bundle/Event/ResourceControllerEvent.php
@@ -37,7 +37,9 @@ class ResourceControllerEvent extends GenericEvent
 
     private ?Response $response = null;
 
-    /** @psalm-suppress MissingReturnType */
+    /**
+     * @psalm-suppress MissingReturnType
+     */
     public function stop(string $message, string $type = self::TYPE_ERROR, array $parameters = [], int $errorCode = 500)
     {
         $this->messageType = $type;
--- a/src/Bundle/Event/ResourceControllerEvent.php
+++ b/src/Bundle/Event/ResourceControllerEvent.php
@@ -37,9 +37,7 @@ class ResourceControllerEvent extends GenericEvent
 
     private ?Response $response = null;
 
-    /**
-     * @psalm-suppress MissingReturnType
-     */
+    /** @psalm-suppress MissingReturnType */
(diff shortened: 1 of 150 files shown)


src/Bundle/Event/ResourceControllerEvent.php
```diff
@@ -37,7 +37,9 @@ class ResourceControllerEvent extends GenericEvent
 
     private ?Response $response = null;
 
-    /** @psalm-suppress MissingReturnType */
+    /**
+     * @psalm-suppress MissingReturnType
+     */
```
GSadee: It looked better before change

src/Bundle/Grid/Parser/OptionsParser.php
```diff
@@ -29,13 +29,16 @@ final class OptionsParser implements OptionsParserInterface
...
-        PropertyAccessorInterface $propertyAccessor
+        PropertyAccessorInterface $propertyAccessor,
     ) {
         $this->container = $container;
         $this->expression = $expression;
         $this->propertyAccessor = $propertyAccessor;
     }
 
+    /**
+     * @param null $data
```
GSadee:
```suggestion
     * @param mixed $data
```
or even better to list explicitly possible types

src/Bundle/Storage/CookieStorage.php
```diff
@@ -58,6 +58,9 @@ public function onKernelResponse(ResponseEvent $event): void
         }
 
         $response = $event->getResponse();
+        /**
+         * @var string|null $value
+         */
```
GSadee: One line?

GSadee (approved):

src/Bundle/Event/ResourceControllerEvent.php
Zales0123: That's... true 😄

PR #461, opened 2022-07-05
[docker]Dockerized Resource Bundle

Ferror:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

## Usage

Build: `docker compose up -d --build`
Execute: `make test`

or run any other command you want via: `docker compose run --rm package your command with parameters`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+ARG COMPOSER_VERSION=2.3
+ARG PHP_VERSION=8.0
+
+FROM composer:${COMPOSER_VERSION} AS composer
+FROM mlocati/php-extension-installer AS php_extension_installer
+
+FROM php:${PHP_VERSION}-cli-alpine AS php
+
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+COPY --from=composer /usr/bin/composer /usr/bin/composer
+COPY --from=php_extension_installer /usr/bin/install-php-extensions /usr/bin/install-php-extensions
+
+RUN install-php-extensions pdo_sqlite
+
+COPY . /app
+
+WORKDIR /app
+
+RUN composer install --no-interaction --no-progress
```
lchrusciel: Won't it be better to have it always up-to-date? We don't store `composer.lock` here anyway
```suggestion
RUN composer update --with-all-dependencies --no-interaction --no-progress
```

lchrusciel (approved):

PR #469, opened 2022-07-27
Return a proper 422 HTTP status code when the form submission fails

belmeopmenieuwesim:
I noticed that the resource controller does not set the 422 http status code when a form submission fails. It is a good practice to do so. Also this will bring support for Symfony UX Turbo, which requires the status code to be set to 422 upon form failure. 

| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Bundle/Controller/ControllerTrait.php
+++ b/src/Bundle/Controller/ControllerTrait.php
@@ -17,6 +17,7 @@ use Psr\Link\LinkInterface;
 use Symfony\Component\Form\Extension\Core\Type\FormType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\Form\FormInterface;
+use Symfony\Component\Form\FormView;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
 use Symfony\Component\HttpFoundation\JsonResponse;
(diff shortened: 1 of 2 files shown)


src/Bundle/Controller/ControllerTrait.php
```diff
@@ -246,6 +247,36 @@ protected function render(string $view, array $parameters = [], Response $respon
...
+            }
+
+            if (!$v instanceof FormInterface) {
+                continue;
+            }
+
+            $parameters[$k] = $v->createView();
+
+            if (200 === $response->getStatusCode() && $v->isSubmitted() && !$v->isValid()) {
+                $response->setStatusCode(422);
```
loic425, line 273: Why not using Resonse::UNPROCESSABLE_ENTITY?

does the code you wrote come from Symfony itself?

belmeopmenieuwesim: Yes i just copied it directly from Symfony's `AbstractController`.

belmeopmenieuwesim:
Please look at this..

Zales0123:
Merged in https://github.com/Sylius/SyliusResourceBundle/pull/488 🖖 Thank you!

PR #487, opened 2022-10-18
Make CsrfTokenManager public

### LINKED ISSUE #477: Symfony 6.1 csrf protection
**Sylius version affected**: v1.10.0-ALPHA.1

**Description**  

The csrf_protected value is configured, but is triggering the following error

CSRF protection is not enabled in your application. Enable it with the "csrf_protection" key in "config/packages/framework.yaml". triggered in https://github.com/Sylius/SyliusResourceBundle/blob/1.10/src/Bundle/Controller/ControllerTrait.php#L395

**Steps to reproduce**  

**Possible Solution**  
Set service "security.csrf.token_manager" to public in the same way as is done for twig?

Zales0123:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/Sylius/SyliusResourceBundle/issues/477
| License         | MIT

We, unfortunately, rely on this service to be public in the `ResourceController` :/ 

I've covered the bug with a test (by adding a "delete" button to the test UI) and added explicit Symfony 6.0 and 6.1 testing in the workflow 🖖 

Thank you, @dannyvw, for reporting this bug! 🏅

### DIFF
--- a/src/Bundle/DependencyInjection/Compiler/CsrfTokenManagerPass.php
+++ b/src/Bundle/DependencyInjection/Compiler/CsrfTokenManagerPass.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
 declare(strict_types=1);
 
 namespace Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler;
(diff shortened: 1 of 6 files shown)


loic425 (approved):

src/Bundle/DependencyInjection/Compiler/CsrfTokenManagerPass.php
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * TODO Remove on sylius/resource-bundle 2.0
+ */
+final class CsrfTokenManagerPass implements CompilerPassInterface
```
lchrusciel, line 22: It would be just the art for the sake of art, but we could try to cover this pass with configuration test as well :P

lchrusciel (approved):

PR #488, opened 2022-10-18
Return 422 status code when the form fails

Zales0123:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | replaces https://github.com/Sylius/SyliusResourceBundle/pull/469
| License         | MIT

I've taken over the https://github.com/Sylius/SyliusResourceBundle/pull/469 PR, added some tests and clean up the code. Thank you, @belmeopmenieuwesim, for the contribution 🖖

### DIFF
--- a/src/Bundle/Controller/ResourceController.php
+++ b/src/Bundle/Controller/ResourceController.php
@@ -212,6 +212,9 @@ class ResourceController
 
             return $this->redirectHandler->redirectToResource($configuration, $newResource);
         }
+        if ($request->isMethod('POST') && $form->isSubmitted() && !$form->isValid()) {
+            $responseCode = Response::HTTP_UNPROCESSABLE_ENTITY;
+        }
 
         if (!$configuration->isHtmlRequest()) {
             return $this->createRestView($configuration, $form, Response::HTTP_BAD_REQUEST);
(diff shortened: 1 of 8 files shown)


src/Bundle/Controller/ResourceController.php
```diff
@@ -212,6 +212,9 @@ public function createAction(Request $request): Response
 
             return $this->redirectHandler->redirectToResource($configuration, $newResource);
         }
+        if ($request->isMethod('POST') && $form->isSubmitted() && !$form->isValid()) {
```
lchrusciel, line 215: Small improvement - we could add a blank line before if and extract the content of the check to the separate method

```suggestion

        if ($this->isFormInValid($request, $form)) {
```

src/Bundle/test/src/Tests/Controller/ScienceBookUiTest.php
```diff
@@ -84,6 +84,28 @@ public function it_allows_creating_a_book(): void
...
+
+        $this->loadFixturesFromFile('science_books.yml');
+
+        $this->client->request('GET', '/science-books/new');
+        $this->client->submitForm('Create', [
+            'science_book[title]' => $newBookTitle,
+            'science_book[author][lastName]' => $newBookAuthorLastName,
+        ]);
+
+        $this->assertResponseCode($this->client->getResponse(), Response::HTTP_UNPROCESSABLE_ENTITY);
```
lchrusciel, line 101: I'm not sure if it is not bc-break and if we didn't send a 400 error. But looking at the code and related PR, I assume we didn't, so I'm fine with this change

lchrusciel (approved):

PR #500, opened 2022-11-19
[Adr] Kernel events vs controllers

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/adr/0000-kernel-events.md
+++ b/adr/0000-kernel-events.md
@@ -1,0 +1,44 @@
+# Kernel events
+
+* Status: new
+* Deciders: @lchrusciel, @loic425, @Zales0123, @soyuka, @dunglas
+
+## Context and Problem Statement
+
+ResourceController has all the actions, a lot of pretty similar code on each action and hard to maintain/test.
+
+## Considered Options
+
+* Use Kernel events
+* Use one controller per action
+
+## PreWriteListener
+
+### With kernel events
+
+```php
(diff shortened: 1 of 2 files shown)


adr/0000-kernel-events.md
```diff
@@ -0,0 +1,70 @@
+# Kernel events
+
+* Status: new
+* Deciders: @lchrusciel, @loic425, @Zales0123
+* External advices: @soyuka, @dunglas, @mtarld
+
+## Context and Problem Statement
+
+ResourceController has all the actions, a lot of pretty similar code on each action and hard to maintain/test.
+and it is not flexible enough in its behavior.
```
lchrusciel, line 10:
```suggestion
Over the years ResourceController become really bloated. It has all the actions, and a lot of pretty similar code on each of them and the default approach to extend its logic is to extend the class. This class has 17 dependencies, which makes it hard to maintain and test. Moreover, some dependencies are not needed for every method call. As a result, this class does not follows SOLID principles and we want to improve its desing
```

lchrusciel (commented):
I've tried to influence the file, but I'm lost with the problem we want to solve with it. I was thinking about DTO support at the beginning, but DTOs are not even mentioned here. The usage of kernel events can be discussed without mentioning one controller per action. Let's make a call tomorrow, discuss it and try to build ADR based on that :)

PR #504, opened 2022-11-25
Operations wih all attributes

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes (1.11)
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Component/Metadata/Operation.php
+++ b/src/Component/Metadata/Operation.php
@@ -1,0 +1,471 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Resource\Metadata;
+
+class Operation
+{
+    public function __construct(
+        protected ?string $name = null,
+        protected ?array $methods = null,
+        protected ?string $path = null,
+        protected ?string $routePrefix = null,
+        protected ?string $controller = null,
+        protected ?string $template = null,
+        protected ?array $repository = null,
+        protected ?array $criteria = null,
+        protected ?array $serializationGroups = null,
+        protected ?string $serializationVersion = null,
+        protected ?array $requirements = null,
+        protected ?array $options = null,
+        protected ?string $host = null,
+        protected ?array $schemes = null,
+        protected ?int $priority = null,
+        protected ?array $vars = null,
+        protected string | array | bool | null $form = null,
+        protected string | array | bool | null $factory = null,
+        protected ?string $section = null,
+        protected ?bool $permission = null,
+        protected ?string $grid = null,
+        protected ?bool $csrfProtection = null,
+        protected string | array | null $redirect = null,
+        protected ?array $stateMachine = null,
+        protected ?string $event = null,
+        protected ?bool $returnContent = null,
+        protected ?string $resource = null,
+        protected ?string $provider = null,
+        protected ?string $processor = null,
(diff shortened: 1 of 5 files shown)


src/Component/spec/Metadata/OperationSpec.php
```diff
@@ -0,0 +1,566 @@
...
+    {
+        $this->withRepository(['method' => 'createListQueryBuilder'])
+            ->getRepository()
+            ->shouldReturn(['method' => 'createListQueryBuilder'])
+        ;
+    }
+
+    function it_could_have_a_repository_with_method_and_arguments(): void
+    {
+        $this->withRepository(['method' => 'createListQueryBuilder', 'arguments' => ['foo' => 'fighters']])
```
loic425, line 124: we should add a withRepositoryMethod and a withRepositoryArguments

src/Component/spec/Metadata/OperationSpec.php
```diff
@@ -0,0 +1,566 @@
...
+    }
+
+    function it_has_no_vars_by_default(): void
+    {
+        $this->getVars()->shouldReturn(null);
+    }
+
+    function it_could_have_vars(): void
+    {
+        $this->withVars([
```
loic425, line 215: maybe it could be better to name it templateVars, but it will be harder for conversions.

src/Component/spec/Metadata/OperationSpec.php
```diff
@@ -0,0 +1,566 @@
...
+    {
+        $this->withForm(BookType::class)
+            ->getForm()
+            ->shouldReturn(BookType::class)
+        ;
+    }
+
+    function it_could_have_a_form_with_type_and_validation_groups(): void
+    {
+        $this->withForm(['type' => BookType::class, 'validation_groups' => 'my_custom_group'])
```
loic425, line 244: we should add a withFormType and a withValidationGroups

src/Component/spec/Metadata/OperationSpec.php
```diff
@@ -0,0 +1,566 @@
...
+    {
+        $this->withRedirect('update')
+            ->getRedirect()
+            ->shouldReturn('update')
+        ;
+    }
+
+    function it_could_have_a_redirection_with_route_and_parameters(): void
+    {
+        $this->withRedirect(['route' => 'app_admin_book_show', 'parameters' => ['id' => 'resource.id']])
```
loic425, line 370: we should add a withRedirectRoute and a withRedirectParameters

src/Component/spec/Metadata/OperationSpec.php
```diff
@@ -0,0 +1,566 @@
...
+    }
+
+    function it_has_no_controller_by_default(): void
+    {
+        $this->getController()->shouldReturn(null);
+    }
+
+    function it_could_have_a_controller(): void
+    {
+        $this->withController(DummyController::class)
```
lchrusciel, line 90: For usage like this, I would find it more packed/easier to get to put the fake classes in the same file as the test file

src/Component/Metadata/Operation.php
```diff
@@ -0,0 +1,474 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Resource\Metadata;
+
+/**
+ * @experimental
+ */
+class Operation
```
lchrusciel, line 19: My generic opinion is that it would be easier to iterate over this feature if it will be split into even smaller groups. Not all 34 properties injected into this class will be used at once and this seems like a straight road into dead/unused attributes in the future. Can we introduce Operation with the first usage? It doesn't need to be feature complete. 

But also, let me review rest of the PRs

PR #510, opened 2022-11-29
Fix declaring repository on resource when it is a service entity repository

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no -->
| Related tickets |
| License         | MIT

It fixes this issue https://github.com/Sylius/SyliusResourceBundle/actions/runs/3575281362/jobs/6011624266#step:11:27

### DIFF
--- a/src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineORMDriver.php
+++ b/src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineORMDriver.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Sylius\Bundle\ResourceBundle\DependencyInjection\Driver\Doctrine;
 
 use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\ServiceRepositoryCompilerPass;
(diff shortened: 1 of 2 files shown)


loic425:
I'm wondering if we can go further, cause we can inject other dependencies into the repository. Maybe we should check if the FQCN is already registered and autowired. If we are in this case, we have nothing to do.

src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineORMDriver.php
```diff
@@ -68,7 +68,12 @@ protected function addRepository(ContainerBuilder $container, MetadataInterface
 
             $genericEntities[] = $entityClass;
         } else {
-            $definition->setArguments([$managerReference, $this->getClassMetadataDefinition($metadata)]);
+            if (is_a($repositoryClass, ServiceEntityRepository::class, true)) {
+                $definition->setArguments([new Reference('doctrine')]);
```
lchrusciel, line 72: Why did you decide on a new doctrine reference instead of `new Reference($metadata->getServiceId('manager'));`?

lchrusciel (approved):

PR #514, opened 2022-11-30
Add resource

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no/yes
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Component/Metadata/Operations.php
+++ b/src/Component/Metadata/Operations.php
@@ -1,0 +1,87 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace  Sylius\Component\Resource\Metadata;
+
+use RuntimeException;
+
+/**
+ * @internal
+ */
+final class Operations implements \IteratorAggregate, \Countable
+{
+    private array $operations = [];
+
+    /**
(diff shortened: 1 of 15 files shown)


src/Component/Metadata/Resource.php
```diff
@@ -0,0 +1,53 @@
...
+#[\Attribute(\Attribute::TARGET_CLASS)]
+final class Resource
+{
+    private ?Operations $operations;
+
+    public function __construct(
+        private ?string $alias = null,
+        ?array $operations = null,
+    ) {
+        $this->operations = null === $operations ? null : new Operations($operations);
```
lchrusciel, line 25: will it be that hard just to expect the Operations class to be passed here? What is blocking us from it? Nested attributes?

src/Component/Metadata/Operations.php
```diff
@@ -0,0 +1,87 @@
...
+        foreach ($this->operations as $i => [$operationName, $operation]) {
+            if ($operationName === $key) {
+                unset($this->operations[$i]);
+
+                return $this;
+            }
+        }
+
+        throw new RuntimeException(sprintf('Could not remove operation "%s".', $key));
+    }
```
lchrusciel, line 70: I'm wondering if we will have use case for it

lchrusciel (approved):

src/Component/Metadata/Resource.php
loic425: That's will be less user-friendly. It's only for DX.

PR #516, opened 2022-12-01
📃 Add section

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Component/Metadata/Section.php
+++ b/src/Component/Metadata/Section.php
@@ -1,0 +1,81 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Resource\Metadata;
+
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
+final class Section
+{
+    private ?Operations $operations;
+
+    public function __construct(
+        private ?string $name = null,
+        private ?string $routePrefix = null,
+        private ?string $templatesDir = null,
+        ?array $operations = null,
+    ) {
+        $this->operations = null === $operations ? null : new Operations($operations);
+    }
+
+    public function getName(): ?string
+    {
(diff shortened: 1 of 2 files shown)


src/Component/Metadata/Section.php
```diff
@@ -0,0 +1,81 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Resource\Metadata;
+
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
+final class Section
```
lchrusciel, line 17: Let's rename it to Routing
```suggestion
final class Routing
```

Calling it a `Section` implies it is a superior concept over the routing, while it is an optional configuration. What is more, I had issues understanding what it configures just by seeing naming. `Routing` seems to be much more readable and straightforward name to me. 

However, we should preserve `section` config, but I would recommend putting it as one of keys under `Routing` attribute

src/Component/Metadata/Section.php
```diff
@@ -0,0 +1,81 @@
...
+{
+    private ?Operations $operations;
+
+    public function __construct(
+        private ?string $name = null,
```
lchrusciel, line 22:
```suggestion
        private ?string $section = null,
```

src/Component/Metadata/Section.php
```diff
@@ -0,0 +1,81 @@
...
+        private ?string $routePrefix = null,
```
lchrusciel, line 23: If we rename the class to `Routing`, `route` prefix here seems to be tautology 
```suggestion
        private ?string $prefix = null,
```

src/Component/Metadata/Section.php
loic425, line 17:
```suggestion
final class OperationGroup
```

loic425, line 17:
```suggestion
final class OperationGroup
```

loic425:
@lchrusciel I think we can close this one... We handle this in Resource attribute directly. WDYT?

PR #517, opened 2022-12-01
Add index operation with minimal data

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT


src/Component/Metadata/Index.php
```diff
@@ -0,0 +1,37 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Resource\Metadata;
+
+/**
+ * @experimental
+ */
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
+final class Index extends HttpOperation implements CollectionOperationInterface
```
lchrusciel, line 20: I'm not convinced that we should refer to this operation different on class and interface level

lchrusciel (approved):

PR #521, opened 2022-12-01
➕ Add create operation with minimal data

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT


src/Component/Metadata/Create.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Resource\Metadata;
+
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
+/**
+ * @experimental
+ */
```
loic425:
```suggestion
/**
 * @experimental
 */
 #[\Attribute(\Attribute::TARGET_CLASS | \Attribute::IS_REPEATABLE)]
```

src/Component/Metadata/CreateOperationInterface.php
```diff
@@ -0,0 +1,23 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Component\Resource\Metadata;
+
+/**
+ * The Operation creates content.
+ *
+ * @experimental
+ */
+interface CreateOperationInterface
```
lchrusciel, line 21: Let's put operation methods here or in separate `OperationInterface`

loic425: This CreateOperationInterface will be important to identify it later. On APIP, we sometimes can referred to the http method, but not in our case, cause we have GET or POST method for a create operation.

But we have some interfaces like this one on APIP too:
* https://github.com/api-platform/core/blob/main/src/Metadata/DeleteOperationInterface.php
* https://github.com/api-platform/core/blob/main/src/Metadata/CollectionOperationInterface.php

lchrusciel: I'm not yet convinced about the need for it if it is going to be empty. But for sure, it is better to have a marking interface instead of some method

src/Component/Metadata/Operation.php
```diff
@@ -13,11 +13,13 @@
...
+/**
+ * @experimental
+ */
+abstract class Operation
 {
     public function __construct(
         protected ?string $name = null,
-        protected ?array $methods = null,
         protected ?string $path = null,
         protected ?string $routePrefix = null,
```
lchrusciel: If we want to separate the HTTP operation from generic operations do we need these two lines? I'm wondering if we can see a direct usage of it

loic425: LOL, I've just seen that and you was reviewing it when I did that refactor :woman_dancing:

src/Component/Metadata/CreateOperationInterface.php
loic425: Could we keep this in mind, and let it as is for now?

src/Component/Metadata/CreateOperationInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Resource\Metadata;
+
+/**
+ * The Operation creates content.
+ *
```
lchrusciel, line 18: I don't see too much value in this comment. Either let's elaborate more or delete it
```suggestion
```

lchrusciel (approved):

src/Component/Metadata/Operation.php
```diff
@@ -13,13 +13,13 @@
...
+ */
+abstract class Operation
 {
     public function __construct(
         protected ?string $name = null,
-        protected ?array $methods = null,
-        protected ?string $path = null,
-        protected ?string $routePrefix = null,
         protected ?string $controller = null,
         protected ?string $template = null,
```
lchrusciel, line 24: I'm wondering if template should be part of operations in general

loic425: Yeah indeed...

PR #525, opened 2022-12-02
🕵️ Remove custom controller

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no but this is advised by Mathias
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

 During the meet with @mtarld, he advised me to remove the controller customization. This introduced a lot of bugs and misuses. We can handle everything with state providers and processors.


mtarld:
For the record, a custom controller is a good feature but is often misused by users.
For example, they often expect listeners to be called but they're not.

So, IMHO you can either remove that feature (if you think that it doesn't really bring anything) or document it very carefully by saying to use it wisely.

lchrusciel:
As I'm still in the camp (perhaps I may be just alone here) that we should preserve the controller as a primary vehicle for the logic execution + provide some DX-improving services and/or traits, I would vote for leaving it. Issues that Matthias mentioned would be mitigated because we are using the controller as well, and therefore someone cannot expect listeners to do something with the content, as they are just jumping right on the front of the whole execution. 

However, if you are convinced that it is a right approach, I may merge it

loic425:
@lchrusciel I think for now, cause we are just on operations, contexts and factories, I think we can remove this controller part for sure. We'll create it when it would be necessary or not.

lchrusciel:
After thinking about it during the night, I totally agree. We can introduce it once it will be needed

lchrusciel (approved):

PR #527, opened 2022-12-05
Add context

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Example of usage
https://github.com/Sylius/SyliusResourceBundle/pull/527#discussion_r1040856381

### DIFF
--- a/src/Component/Context/Context.php
+++ b/src/Component/Context/Context.php
@@ -1,0 +1,79 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Resource\Context;
+
+/**
+ * @implements \IteratorAggregate<object>
+ */
+final class Context implements \IteratorAggregate
+{
+    /** @var array<class-string, object> */
+    private array $optionMap;
+
+    public function __construct(object ...$options)
+    {
+        $map = [];
+        foreach ($options as $option) {
+            $map[get_class($option)] = $option;
+        }
+
+        $this->optionMap = $map;
+    }
+
+    public function with(object ...$options): self
+    {
+        /** @psalm-suppress DuplicateArrayKey */
+        return new self(...[
+            ...array_values($this->optionMap),
+            ...$options,
+        ]);
+    }
+
+    /**
+     * @param class-string $optionClasses
+     */
+    public function without(string ...$optionClasses): self
+    {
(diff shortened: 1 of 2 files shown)


src/Component/Context/Context.php
```diff
@@ -0,0 +1,85 @@
...
+final class Context implements \IteratorAggregate
+{
+    /** @var array<class-string, object> */
+    private array $optionMap;
+
+    public function __construct(object|array ...$options)
+    {
+        $map = [];
+        foreach ($options as $option) {
+            if (\is_array($option)) {
```
loic425: @mtarld I've improved that part to pass an interface as a key instead of the class.
WDYT?

mtarld: Just wondering, why do you need to have an interface as a key there?
Be careful btw as it might introduce collisions

loic425: Not needed now, as discussed together.

loic425: Encapsulate the service in a context option seems to be perfect.

```php
final class InputOption
{
    public function __construct(private InputInterface $input)
    {
    }

    public function input(): InputInterface
    {
        return $this->input;
    }
}
```
Example of its usage
```php
$inputOption = $context->get(InputOption::class);
$outputOption = $context->get(OutputOption::class);

Assert::notNull($inputOption);
Assert::notNull($outputOption);

$ui = new SymfonyStyle($inputOption->input(), $outputOption->output());
```

src/Component/Context/Context.php
```diff
@@ -0,0 +1,79 @@
...
+     * @template T of object
+     *
+     * @param class-string<T> $optionClass
+     *
+     * @return T|null
+     */
+    public function get(string $optionClass): ?object
+    {
+        /** @var T $option */
+        $option = $this->optionMap[$optionClass] ?? null;
```
lchrusciel, line 67: what is the case that we are ok with receiving null here? Won't it be better that `get` would ensure the existence of the requested option?

lchrusciel (approved):

src/Component/Context/Context.php
lchrusciel: I'm just wondering why won't we prefer to have

```php
$inputOption = $context->get(InputOption::class);
$outputOption = $context->get(OutputOption::class);

$ui = new SymfonyStyle($inputOption->input(), $outputOption->output());
```

with still being sure, that these classes exists :)

loic425: Yes but not be sure it is in the context. Cause it depends to the... Context (cli or http for exemple)

PR #534, opened 2022-12-13
👬 Resource metadata collection factory

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

The whole code is a lot inspired by https://github.com/api-platform/core/blob/main/src/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php

**One attribute per line**

```php
#[Resource(alias: 'app.book', section: 'admin')]
#[Index]
#[Create]
#[Resource(alias: 'app.book', section: 'shop')]
#[Index]
#[Show]
final class BookResource
{
}
```

**Nested operations**

```php
#[Resource(
    alias: 'app.book', 
    section: 'admin',
    operations: [
        new Index(),
        new Create(),
    ],
)]
#[Resource(
    alias: 'app.book', 
    section: 'shop',
    operations: [
        new Index(),
        new Show(),
    ],
)]
final class BookResource
{
}
```

**Without resource attribute**

On this next example, it will search for a resource with `getByClass` method on the resource registry.

```php
#[Index]
#[Create]
final class BookResource
{
}
```

### DIFF
--- a/src/Component/Metadata/Operations.php
+++ b/src/Component/Metadata/Operations.php
@@ -41,7 +41,7 @@ final class Operations implements \IteratorAggregate, \Countable
         })();
     }
 
-    public function get(string $key): ?Operation
+    public function get(string $key): Operation
     {
         foreach ($this->operations as [$operationName, $operation]) {
             if ($operationName === $key) {
@@ -49,7 +49,7 @@ final class Operations implements \IteratorAggregate, \Countable
             }
         }
 
-        return null;
+        throw new \RuntimeException(sprintf('No Operation with key "%s" was found', $key));
     }
 
     public function add(string $key, Operation $value): self
--- a/src/Component/Metadata/Operations.php
+++ b/src/Component/Metadata/Operations.php
@@ -41,6 +41,17 @@ final class Operations implements \IteratorAggregate, \Countable
         })();
     }
 
+    public function get(string $key): ?Operation
+    {
+        foreach ($this->operations as [$operationName, $operation]) {
+            if ($operationName === $key) {
+                return $operation;
+            }
+        }
+
+        return null;
+    }
+
     public function add(string $key, Operation $value): self
     {
         foreach ($this->operations as $i => [$operationName, $operation]) {
--- a/src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
+++ b/src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
@@ -14,18 +14,21 @@ declare(strict_types=1);
 namespace Sylius\Component\Resource\Metadata\Resource\Factory;
 
 use Sylius\Component\Resource\Metadata\HttpOperation;
+use Sylius\Component\Resource\Metadata\MetadataInterface;
 use Sylius\Component\Resource\Metadata\Operation;
 use Sylius\Component\Resource\Metadata\Operations;
 use Sylius\Component\Resource\Metadata\RegistryInterface;
-use Sylius\Component\Resource\Metadata\Resource;
 use Sylius\Component\Resource\Metadata\Resource as ResourceMetadata;
 use Sylius\Component\Resource\Metadata\Resource\ResourceMetadataCollection;
 use Sylius\Component\Resource\Reflection\ClassReflection;
+use Sylius\Component\Resource\Symfony\Routing\Factory\OperationRouteNameFactory;
 
 final class AttributesResourceMetadataCollectionFactory implements ResourceMetadataCollectionFactoryInterface
 {
-    public function __construct(private RegistryInterface $resourceRegistry)
-    {
+    public function __construct(
+        private RegistryInterface $resourceRegistry,
+        private OperationRouteNameFactory $operationRouteNameFactory,
+    ) {
     }
 
     public function create(string $resourceClass): ResourceMetadataCollection
@@ -48,7 +51,7 @@ final class AttributesResourceMetadataCollectionFactory implements ResourceMetad
      */
     private function buildResourceOperations(array $attributes, string $resourceClass): array
     {
-        /** @var ResourceMetadata[] $resources */
+        /** @var array<int, ResourceMetadata> $resources */
         $resources = [];
         $index = -1;
 
@@ -73,9 +76,12 @@ final class AttributesResourceMetadataCollectionFactory implements ResourceMetad
             }
 
             if (null === ($resources[$index] ?? null)) {
-                $metadata = $this->resourceRegistry->getByClass($resourceClass);
+                $resourceConfiguration = $this->resourceRegistry->getByClass($resourceClass);
 
-                $resources[++$index] = new Resource($metadata->getAlias());
+                $resource = new ResourceMetadata($resourceConfiguration->getAlias());
+                $resource = $this->getResourceWithDefaults($resource, $resourceConfiguration);
+
+                $resources[++$index] = $resource;
             }
 
             if (!is_subclass_of($attribute->getName(), Operation::class)) {
@@ -96,15 +102,36 @@ final class AttributesResourceMetadataCollectionFactory implements ResourceMetad
         return $resources;
     }
 
+    private function getResourceWithDefaults(ResourceMetadata $resource, MetadataInterface $resourceConfiguration): ResourceMetadata
+    {
+        if (null === $resource->getApplicationName()) {
+            $resource = $resource->withApplicationName($resourceConfiguration->getApplicationName());
+        }
+
+        if (null === $resource->getName()) {
+            $resource = $resource->withName($resourceConfiguration->getName());
+        }
+
+        return $resource;
+    }
+
     private function getOperationWithDefaults(ResourceMetadata $resource, Operation $operation): array
     {
-        if (null !== $section = $resource->getSection()) {
-            $operation = $operation->withSection($section);
+        $resourceConfiguration = $this->resourceRegistry->get($resource->getAlias());
+
+        if (null === $resource->getApplicationName()) {
+            $resource = $resource->withApplicationName($resourceConfiguration->getApplicationName());
         }
+
+        if (null === $resource->getName()) {
+            $resource = $resource->withName($resourceConfiguration->getName());
+        }
+
+        $operation = $operation->withResource($resource);
 
         if ($operation instanceof HttpOperation) {
             if (null === $routeName = $operation->getRouteName()) {
-                $routeName = $this->getDefaultRouteName($resource, $operation);
+                $routeName = $this->operationRouteNameFactory->createRouteName($operation);
                 $operation = $operation->withRouteName($routeName);
             }
 
@@ -114,20 +141,5 @@ final class AttributesResourceMetadataCollectionFactory implements ResourceMetad
         $operationName = $operation->getName();
 
         return [$operationName, $operation];
-    }
-
-    private function getDefaultRouteName(ResourceMetadata $resource, HttpOperation $operation): string
-    {
-        $resourceConfiguration = $this->resourceRegistry->get($resource->getAlias());
-
-        if (null !== $section = $operation->getSection()) {
-            $section = '_' . $section;
-        }
-
-        if (null === $shortName = $operation->getShortName()) {
-            throw new \RuntimeException(sprintf('Operation "%s" should have a short name to build a route name', $operation::class));
-        }
-
-        return sprintf('%s%s_%s_%s', $resourceConfiguration->getApplicationName(), $section ?? '', $resourceConfiguration->getName(), $shortName);
     }
 }
--- a/src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
+++ b/src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
@@ -17,6 +17,7 @@ use Sylius\Component\Resource\Metadata\HttpOperation;
 use Sylius\Component\Resource\Metadata\Operation;
 use Sylius\Component\Resource\Metadata\Operations;
 use Sylius\Component\Resource\Metadata\RegistryInterface;
+use Sylius\Component\Resource\Metadata\Resource;
 use Sylius\Component\Resource\Metadata\Resource as ResourceMetadata;
 use Sylius\Component\Resource\Metadata\Resource\ResourceMetadataCollection;
 use Sylius\Component\Resource\Reflection\ClassReflection;
@@ -72,7 +73,9 @@ final class AttributesResourceMetadataCollectionFactory implements ResourceMetad
             }
 
             if (null === ($resources[$index] ?? null)) {
-                throw new \RuntimeException(sprintf('No Resource attribute was found on %s', $resourceClass));
+                $metadata = $this->resourceRegistry->getByClass($resourceClass);
+
+                $resources[++$index] = new Resource($metadata->getAlias());
             }
 
             if (!is_subclass_of($attribute->getName(), Operation::class)) {
--- a/src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
+++ b/src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
@@ -1,0 +1,82 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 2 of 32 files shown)


src/Component/Metadata/Operations.php
```diff
@@ -41,6 +41,17 @@ public function getIterator(): \Traversable
         })();
     }
 
+    public function get(string $key): ?Operation
+    {
+        foreach ($this->operations as [$operationName, $operation]) {
+            if ($operationName === $key) {
+                return $operation;
+            }
+        }
```
lchrusciel, line 50: Why don't we use `operationName` as a key in the dictionary? Right now, we are creating an ordered list and iterating over items to find proper operation

src/Component/Metadata/Operations.php
```diff
@@ -41,6 +41,17 @@ public function getIterator(): \Traversable
         })();
     }
 
+    public function get(string $key): ?Operation
```
lchrusciel: Do we want to support the case when the operation is null? What is the use case for that?

src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,82 @@
...
+            $resourceMetadataCollection[] = $resource;
+        }
+
+        return $resourceMetadataCollection;
+    }
+
+    /**
+     * @param \ReflectionAttribute[] $attributes
+     */
+    private function buildResourceOperations(array $attributes, string $resourceClass): array
```
lchrusciel:
```suggestion
    /**
     * @param \ReflectionAttribute[] $attributes
     *
     * @return ResourceMetadata[]
     */
    private function buildResourceOperations(array $attributes, string $resourceClass): array
```
But perhaps the `ResourceMetadata[]` could be even translated to custom array object. WDYT?

src/Component/Tests/Dummy/DummyMultiResourcesWithOperations.php
```diff
@@ -0,0 +1,29 @@
...
+namespace Sylius\Component\Resource\Tests\Dummy;
+
+use Sylius\Component\Resource\Metadata\Create;
+use Sylius\Component\Resource\Metadata\Index;
+use Sylius\Component\Resource\Metadata\Resource;
+use Sylius\Component\Resource\Metadata\Show;
+
+#[Resource(alias: 'app.order')]
+#[Index]
+#[Create]
```
lchrusciel, line 23: Seeing these attributes, got me and the Sylius team thinking if we should allow for such a structure. I would vote (based on the feedback from @Zales0123 and @coldic3) to always require `Routing` attribute from https://github.com/Sylius/SyliusResourceBundle/pull/516/files and not allow operation declaration on the resource

src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,82 @@
...
+        foreach ($attributes as $attribute) {
+            if (is_a($attribute->getName(), ResourceMetadata::class, true)) {
+                /** @var ResourceMetadata $resource */
+                $resource = $attribute->newInstance();
+                $resources[++$index] = $resource;
+
+                continue;
+            }
+
+            Assert::notNull($resources[$index] ?? null, 'No Resource attribute was found');
```
lchrusciel: Two issues with this code:
1. This will always be triggered when the attribute ResourceMetadata won't be attached to `resources` during first iteration. This means that the whole content below can be just placed in foreach loop. TBH, I don't get the idea of index field usage at all. 
 
Looking at the tests, we are missing the case, where we have any other attribute above the `Resource` defined in class. I'm quite sure, that it would fail the current generation logic

src/Component/Metadata/Resource/ResourceMetadataCollection.php
```diff
@@ -0,0 +1,34 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Component\Resource\Metadata\Resource;
+
+use Sylius\Component\Resource\Metadata\Operation;
+use Sylius\Component\Resource\Metadata\Resource as ResourceMetadata;
+
+final class ResourceMetadataCollection extends \ArrayObject
+{
+    public function getOperation(string $resourceAlias, string $name): ?Operation
```
lchrusciel: Do we have a case for null as return here?
```suggestion
    public function getOperation(string $resourceAlias, string $name): Operation
```

src/Component/Metadata/Resource/ResourceMetadataCollection.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /** @var ResourceMetadata $current */
+        foreach ($this->getIterator() as $current) {
+            if ($current->getAlias() !== $resourceAlias) {
```
lchrusciel: Why don't we treat alias as a key in the array?

src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,82 @@
...
+final class AttributesResourceMetadataCollectionFactory implements ResourceMetadataCollectionFactoryInterface
+{
+    public function create(string $resourceClass): ResourceMetadataCollection
+    {
+        $resourceMetadataCollection = new ResourceMetadataCollection();
+
+        $attributes = ClassReflection::getClassAttributes($resourceClass);
+
+        foreach ($this->buildResourceOperations($attributes, $resourceClass) as $resource) {
+            $resourceMetadataCollection[] = $resource;
```
lchrusciel, line 41: Can we adjust this logic to following rules:
1. One class may have several resource attributes
2. Each resource attribute requires alias and this alias is unique among all resources
3. There is no possibility to define any operations on your own
4. To declare any operation, you have to add `Routing` attribute, which will require alias field to be declared and matched with any resource alias (the question is, should we allow to define routing for aliases from other classes)
To improve DX in the next iteration we can consider autogeneration of alias with the pattern: `app.your_resource`

loic425:
@lchrusciel I think about all these rules and feedbacks and I think I have sth to present you and that we can both agree.
I'll anwser on this PR when I'll have enough time to explain all my ideas and rules.

src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
loic425: > Can we adjust this logic to following rules:
> 
>     1. One class may have several resource attributes

agree

> 
>     2. Each resource attribute requires alias and this alias is unique among all resources

for the required alias, I agree. 
For the unique alias, I'm not sure, I think we can use the same alias on an extended class to customize already defined operations in sub class.

> 
>     3. There is no possibility to define any operations on your own

Why? I think it's very interesting to allow them.

> 
>     4. To declare any operation, you have to add `Routing` attribute, which will require alias field to be declared and matched with any resource alias (the question is, should we allow to define routing for aliases from other classes)

For me, you always need a resource attribute to define operations. But the one attribute per line is interesting. Moreover, one of the main goal of this implementation is to be closer to API Platform. Thus, Sylius will be easier for beginners to learn. For now, a lot of Symfony users are struggling with the resource system.

>        To improve DX in the next iteration we can consider autogeneration of alias with the pattern: `app.your_resource`

Yep, I'd like too, I agree.

src/Component/Metadata/Operations.php
loic425: If we don't allow null here, we have to throw a runtime exception, yes that seems good

src/Component/Metadata/Operations.php
loic425: ;that's a lot copied/pasted from APIP, but indeed I think we can improve a lot by using the operation name as the key

src/Component/Tests/Dummy/DummyMultiResourcesWithOperations.php
loic425: :sob:

src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
loic425: Yeah, I have to improve that. For me, as this is necessary to define a resource attribute (for my point of view), We have to inform the developer about it.

With Entity attribute for example, when we extend an entity, we have to redefine the entity attribute too.

src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
loic425: @lchrusciel Look at the first post where I put some examples

src/Component/Metadata/Resource/ResourceMetadataCollection.php
loic425: Yes, that's a good idea

src/Component/Metadata/Resource/ResourceMetadataCollection.php
loic425: A runtime exception will be better

src/Component/Metadata/Operations.php
lchrusciel: I'm reviewing it as our code. Do we want to keep it coherent 1-to-1 with APIP to one day integrate it? I would prefer simplify it, so it will be easier to maintain :)

src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
lchrusciel: 2. Each resource attribute requires alias and this alias is unique among all resources
   for the required alias, I agree. For the unique alias, I'm not sure, I think we can use the same alias on an extended class to customize already defined operations in sub class.

agree :)

3. There is no possibility to define any operations on your own
   Why? I think it's very interesting to allow them.
After deeper thinking, you are correct - we allow for them with custom route definition in Sylius :+1:

src/Component/Metadata/Resource/ResourceMetadataCollection.php
```diff
@@ -0,0 +1,42 @@
...
+final class ResourceMetadataCollection extends \ArrayObject
+{
+    public function getOperation(string $resourceAlias, string $name, ?string $section = null): Operation
+    {
+        /** @var ResourceMetadata $current */
+        foreach ($this->getIterator() as $current) {
+            if (
+                $current->getAlias() !== $resourceAlias ||
+                $section !== $current->getSection() ||
+                null === ($operation = $current->getOperations()?->get($name))
```
lchrusciel: I'm picky, but this logic of replaced order of `current` variable and value to comapare is annoying me. Can we fix it?
```suggestion
                $current->getAlias() !== $resourceAlias ||
                $current->getSection() !== $section ||
```

lchrusciel: WDYT? For me it is much more readable
```suggestion
!$current->hasOperations($name)
```

src/Component/Metadata/Resource/ResourceMetadataCollection.php
```diff
@@ -0,0 +1,42 @@
...
+            ) {
+                continue;
+            }
+
+            return $operation;
+        }
```
lchrusciel: if we will revert the conditions in `if` statement we can go with

```suggestion
            ) {
                return $operation;
            }
```

which is simpler in my opinion :) WDYT?

src/Component/Metadata/Resource/ResourceMetadataCollection.php
```diff
@@ -0,0 +1,42 @@
...
+
+        throw new \RuntimeException(sprintf(
+            'Operation "%s" for "%s" resource was not found.',
```
lchrusciel, line 34: According to the `if` we should add section as well

src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
lchrusciel: TBH, perhaps we should discuss just cases, that we want to cover and simplify the code to support only them. What would you say?

src/Component/Tests/Dummy/DummyResourceWithSections.php
```diff
@@ -0,0 +1,29 @@
...
+use Sylius\Component\Resource\Metadata\Index;
+use Sylius\Component\Resource\Metadata\Resource;
+use Sylius\Component\Resource\Metadata\Show;
+
+#[Resource(alias: 'app.dummy', section: 'admin')]
+#[Index]
+#[Create]
+#[Resource(alias: 'app.dummy', section: 'shop')]
+#[Index]
+#[Show]
```
lchrusciel, line 26: I'm still thinking how should we handle this. Even tho, I'm on a boat with the idea of having `#[Index]` as separate property, I'm still not convinced on resolution of them just by its placement under the resource. Can we discuss this design decision

src/Component/Metadata/Resource/ResourceMetadataCollection.php
loic425: That's pretty much more readable, but we also have to get the operation. But I will commit your suggested changes and adjust them.

src/Component/Tests/Dummy/DummyResourceWithSections.php
loic425: Yeah good idea, we have to find a solution that we are both proud of first.

src/Component/Metadata/Resource/ResourceMetadataCollection.php
loic425: done

src/Component/Metadata/Resource/ResourceMetadataCollection.php
loic425: There was a conflict on suggested changes. I've tried to do my best, but you have to check again, sorry.

src/Component/Metadata/Operations.php
loic425: done

src/Component/Tests/Dummy/DummyMultiResourcesWithOperations.php
loic425: Could we resolve that? It's also necessary if we want to create operations without adding a resource attribute for a resource which is already defined on the Resource bundle configuration.

src/Component/Tests/Dummy/DummyResourceWithSections.php
loic425: Do you have a proposal?

src/Component/Metadata/Resource/ResourceMetadataCollection.php
loic425: Not necessary now, cause we don't have section anymore here. We handle name with the route name of the operation when it's an http operation. For CLI operation, we'll handle this with the operation name too.

src/Component/Metadata/Resource/ResourceMetadataCollection.php
loic425: Now, the operation name equals the route name when it's an http operation. 
That's not easy cause the operation name is a concatenation and it's built later.
That's hard to build it earlier when we build the Operations via the nested operations.

```php
#[Resource(
    alias: 'app.dummy',
    section: 'admin',
    operations: [
        new Index(),
        new Create(),
    ],
)]
```

```php
final class Resource
{
    private ?Operations $operations;

    public function __construct(
        private string $alias,
        private ?string $section = null,
        ?array $operations = null,
    ) {
        $this->operations = null === $operations ? null : new Operations($operations);
    }
```

```php
final class Operations implements \IteratorAggregate, \Countable
{
    private array $operations = [];

    public function __construct(array $operations = [])
    {
        foreach ($operations as $operationName => $operation) {
            $this->operations[] = [$operationName, $operation];
        }
    }

    // [...]
}
```

src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
loic425: I was inspired by the 3.0 APIP and it has been improved. 
I decided to try using the resource definition of the class and that works.

src/Component/Metadata/Operations.php
loic425: Of course, you're right.

See this comment https://github.com/Sylius/SyliusResourceBundle/pull/534#discussion_r1069462990

lchrusciel (approved):

PR #549, opened 2023-01-03
Hotfix for Attributes routing system

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

PR #498, opened 2022-11-14
Add services' aliases to improve autowiring experience

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

PR #551, opened 2023-01-03
Flip back service ids and FQCN

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

Replace #549 

Fix the BC-break from #498

### DIFF
--- a/src/Bundle/Resources/config/services.xml
+++ b/src/Bundle/Resources/config/services.xml
@@ -37,20 +37,20 @@
         <service id="sylius.resource_registry" class="Sylius\Component\Resource\Metadata\Registry" public="false" />
         <service id="Sylius\Component\Resource\Metadata\RegistryInterface" alias="sylius.resource_registry" public="false" />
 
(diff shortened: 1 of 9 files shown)


mbabker (approved):

loic425:
@mbabker During internal discussions, we suggest to duplicate the services (both with IDs and FQCN) and adding deprecations to the IDs ones. WDYT?

loic425:
@mbabker Exactly what I did here for only 2 services for now #549

mbabker:
I think that'd be a little overkill.  At that point, you have two independent services and an asinine number of logged container deprecations that you can't do anything about.

The closest that Symfony suggests to a naming convention for service IDs in bundles in the [best practices](https://symfony.com/doc/current/bundles/best_practices.html#services) is prefixing service IDs.  That to me implies that it's better for bundles to not use FQCN's as service IDs, and to purposefully create aliases for the FQCN of interfaces that a service defines where it's feasible.

GSadee (approved):

GSadee:

PR #556, opened 2023-01-13
🛣️ Add operation route factory

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    |yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Based on #534 

Now we can build the route itself :dancers:

### DIFF
--- a/src/Bundle/Routing/OperationRouteFactory.php
+++ b/src/Bundle/Routing/OperationRouteFactory.php
@@ -1,0 +1,91 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ResourceBundle\Routing;
+
+use Gedmo\Sluggable\Util\Urlizer;
+use Sylius\Component\Resource\Metadata\CreateOperationInterface;
+use Sylius\Component\Resource\Metadata\HttpOperation;
+use Sylius\Component\Resource\Metadata\MetadataInterface;
+use Sylius\Component\Resource\Metadata\Resource;
+use Sylius\Component\Resource\Metadata\UpdateOperationInterface;
+use Symfony\Component\Routing\Route;
+
+final class OperationRouteFactory implements OperationRouteFactoryInterface
+{
+    public function create(MetadataInterface $metadata, Resource $resource, HttpOperation $operation): Route
+    {
+        $routePath = $operation->getPath() ?? $this->getDefaultRoutePath($metadata, $operation);
+
+        if (null !== $routePrefix = $operation->getRoutePrefix()) {
+            $routePath = $routePrefix . '/' . $routePath;
+        }
+
+        return new Route(
+            path: $routePath,
+            defaults: [
+                '_sylius' => $this->getSyliusOptions($resource, $operation),
+            ],
+            methods: $operation->getMethods() ?? [],
+        );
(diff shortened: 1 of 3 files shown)


src/Bundle/spec/Routing/OperationRouteFactorySpec.php
```diff
@@ -0,0 +1,200 @@
...
+            new Resource('app.dummy'),
+            new Show(template: 'dummy/show.html.twig'),
+        );
+
+        $route->getPath()->shouldReturn('/dummies/{id}');
+        $route->getMethods()->shouldReturn(['GET']);
+        $route->getDefaults()->shouldReturn([
+            '_sylius' => [
+                'resource' => 'app.dummy',
+                'template' => 'dummy/show.html.twig'
```
loic425: I'm wondering if we really need this, it's a lot simpler for now due to the existing system, but later we can directly have these information from the Operation itself and it will be a lot simpler route.

src/Bundle/Routing/OperationRouteFactory.php
```diff
@@ -0,0 +1,100 @@
...
+
+    private function getDefaultRoutePathForOperation(MetadataInterface $metadata, HttpOperation $operation): string
+    {
+        $rootPath = sprintf('%s', Urlizer::urlize($metadata->getPluralName()));
+
+        if (null !== $path = $operation->getPath()) {
+            return $path;
+        }
+
+        $shortName = $operation->getShortName();
```
loic425: an error should be throw when short name is null.

src/Bundle/Routing/OperationRouteFactory.php
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ResourceBundle\Routing;
```
loic425, line 14: I think that's be better to move this one to Symfony\Component\Resource\Symfony\Routing but we can do that later.

src/Bundle/Routing/OperationRouteFactory.php
```diff
@@ -0,0 +1,91 @@
...
+            methods: $operation->getMethods() ?? [],
+        );
+    }
+
+    private function getDefaultRoutePath(MetadataInterface $metadata, HttpOperation $operation): string
+    {
+        return $this->getDefaultRoutePathForOperation($metadata, $operation);
+    }
+
+    private function getDefaultRoutePathForOperation(MetadataInterface $metadata, HttpOperation $operation): string
```
lchrusciel, line 48: Would be good to extract this logic to separate class. Or even set of classes

src/Bundle/Routing/OperationRouteFactory.php
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        if ('show' === $shortName) {
+            return sprintf('%s/{id}', $rootPath);
+        }
+
+        throw new \InvalidArgumentException(sprintf('Impossible to get a default route path for this operation "%s". Please define a path.', $operation::class));
+    }
+
+    private function getSyliusOptions(Resource $resource, HttpOperation $operation): array
```
lchrusciel, line 87: Same here

lchrusciel (approved):

src/Bundle/Routing/OperationRouteFactory.php
loic425: Yes, it's on a next PR :)

PR #557, opened 2023-01-16
🛣️ Add attributes operation route factory

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Based on #556 

All these services could be moved into `Sylius\Component\Resource\Symfony\Routing` namespace.

### DIFF
--- a/src/Bundle/Routing/AttributesOperationRouteFactoryInterface.php
+++ b/src/Bundle/Routing/AttributesOperationRouteFactoryInterface.php
@@ -1,0 +1,22 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ResourceBundle\Routing;
+
+use Symfony\Component\Routing\RouteCollection;
+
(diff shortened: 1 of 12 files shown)


src/Bundle/Routing/AttributesOperationRouteFactoryInterface.php
```diff
@@ -0,0 +1,22 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ResourceBundle\Routing;
+
+use Symfony\Component\Routing\RouteCollection;
+
+interface AttributesOperationRouteFactoryInterface
+{
+    /** @psalm-param class-string $className */
+    public function createRouteForClass(RouteCollection $routeCollection, string $className): void;
```
lchrusciel, line 21: WDYT about dropping the first argument, returning `RouteCollection` and merging it in the class that calls it? I would say, that it would look better and be more readable. Right now, we are passing route collection only to mutate it at the end
```suggestion
    public function createRouteForClass(string $className): RouteCollection;
```

lchrusciel (approved):

src/Bundle/Routing/AttributesOperationRouteFactoryInterface.php
loic425: Hum, I think I know why it's done like this, but I will check that.

PR #558, opened 2023-01-17
🔨 Add http operation initiator

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

based on #557

### DIFF
--- a/src/Component/Metadata/MetadataInterface.php
+++ b/src/Component/Metadata/MetadataInterface.php
@@ -40,16 +40,18 @@ interface MetadataInterface
     public function getParameter(string $name);
 
     /**
+     * @return class-string $name
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function getClass(string $name): string;
+
+    /**
      * Return all the metadata parameters.
      */
     public function getParameters(): array;
 
     public function hasParameter(string $name): bool;
-
-    /**
-     * @throws \InvalidArgumentException
-     */
-    public function getClass(string $name): string;
 
     public function hasClass(string $name): bool;
 
--- a/src/Component/Metadata/Operation/HttpOperationInitiator.php
+++ b/src/Component/Metadata/Operation/HttpOperationInitiator.php
@@ -1,0 +1,55 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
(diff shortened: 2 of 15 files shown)


src/Component/Metadata/MetadataInterface.php
```diff
@@ -39,18 +39,20 @@ public function getTemplatesNamespace(): ?string;
      */
     public function getParameter(string $name);
 
+    /**
+     * @return class-string $name
```
lchrusciel, line 43: shouldn't it even be `@return class-string $name of ResourceInterface`?

src/Component/Metadata/Operation/HttpOperationInitiator.php
```diff
@@ -0,0 +1,55 @@
...
+            null === $operationName
+        ) {
+            return null;
+        }
+
+        if (str_contains($resource, '.')) {
+            $metadata = $this->resourceRegistry->get($resource);
+        } else {
+            $metadata = $this->resourceRegistry->getByClass($resource);
+        }
```
lchrusciel, line 46: could be extracted to private method to utilize early return instead of else

src/Component/Metadata/Operation/HttpOperationInitiator.php
```diff
@@ -0,0 +1,55 @@
...
+            null === ($resource = $attributes['resource'] ?? null) ||
+            null === $operationName
+        ) {
+            return null;
+        }
+
+        if (str_contains($resource, '.')) {
+            $metadata = $this->resourceRegistry->get($resource);
+        } else {
+            $metadata = $this->resourceRegistry->getByClass($resource);
```
lchrusciel, line 45: this behaviour is not covered by spec

lchrusciel (approved):

src/Component/Metadata/Operation/HttpOperationInitiator.php
loic425: Yeah, I'd like to move this logic on resource registry itself, WDYT?

PR #569, opened 2023-01-23
📝 Add Form listener

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

### DIFF
--- a/src/Component/Metadata/Index.php
+++ b/src/Component/Metadata/Index.php
@@ -30,6 +30,8 @@ final class Index extends HttpOperation implements CollectionOperationInterface
         string|callable|null $processor = null,
         ?bool $read = null,
         ?bool $write = null,
+        ?string $formType = null,
+        ?array $formOptions = null,
     ) {
         parent::__construct(
             methods: $methods ?? ['GET'],
@@ -42,6 +44,8 @@ final class Index extends HttpOperation implements CollectionOperationInterface
             processor: $processor,
             read: $read,
             write: $write,
+            formType: $formType,
+            formOptions: $formOptions,
         );
     }
 }
--- a/src/Component/Symfony/EventListener/FormListener.php
+++ b/src/Component/Symfony/EventListener/FormListener.php
@@ -11,44 +11,41 @@
 
 declare(strict_types=1);
 
-namespace Sylius\Bundle\ResourceBundle\EventListener;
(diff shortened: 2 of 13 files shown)


src/Component/Metadata/Index.php
```diff
@@ -30,6 +30,8 @@ public function __construct(
         string|callable|null $processor = null,
         ?bool $read = null,
         ?bool $write = null,
+        ?string $formType = null,
+        ?array $formOptions = null,
```
lchrusciel, line 34: WDYT about grouping some parts of these attributes into groups and defining nested attributes. Would it be possible to define some Form attribute and interface for it? So if anyone would like to change something, will it be possible? But TBH, the main reason behind this question is that once you would like to extend APIP attributes and you have 30 arguments long constructor it makes you a little bit sad

src/Component/Symfony/EventListener/FormListener.php
```diff
@@ -0,0 +1,51 @@
...
+            null === $operation ||
+            null == $operation->getFormType()
+        ) {
+            return;
+        }
+
+        $form = $this->formFactory->create($operation, $context, $controllerResult);
+        $form->handleRequest($request);
+
+        $request->attributes->set('form', $form);
```
lchrusciel, line 49: Instead of making request attributes our data holder directly, won't it be better to have some additional object to hold this kind of metadata?

src/Component/Symfony/EventListener/FormListener.php
```diff
@@ -0,0 +1,51 @@
...
+        if (
+            $controllerResult instanceof Response ||
+            null === $operation ||
+            null == $operation->getFormType()
+        ) {
+            return;
+        }
+
+        $form = $this->formFactory->create($operation, $context, $controllerResult);
+        $form->handleRequest($request);
```
lchrusciel, line 47: I still believe this logic should be part of controller, but I don't have resources to refactor it, so I should probably remain silent

lchrusciel (approved):

src/Component/Symfony/EventListener/FormListener.php
loic425: Yeah,  I think it's a good idea

PR #579, opened 2023-01-30
Configure templates dir on resource

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

```php
#[Resource(
    alias: 'app.book',
    section: 'admin',
    templatesDir: 'book',
    operations: [
        new Create(
            //template: 'book/create.html.twig',
            provider: Provider::class,
            responder: BookCreationResponder::class,
        ),
    ],
)]
```

### DIFF
--- a/src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
+++ b/src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
@@ -122,6 +122,12 @@ final class AttributesResourceMetadataCollectionFactory implements ResourceMetad
     {
         $resourceConfiguration = $this->resourceRegistry->get($resource->getAlias());
 
+        if (null === $operation->getTemplate()) {
+            $templateDir = $resource->getTemplatesDir() ?? '';
+            $template = sprintf('%s/%s.html.twig', $templateDir, $operation->getShortName() ?? '');
+            $operation = $operation->withTemplate($template);
+        }
+
         if (null === $resource->getApplicationName()) {
             $resource = $resource->withApplicationName($resourceConfiguration->getApplicationName());
         }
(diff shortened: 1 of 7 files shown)


src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
```diff
@@ -122,6 +122,12 @@ private function getOperationWithDefaults(ResourceMetadata $resource, Operation
     {
         $resourceConfiguration = $this->resourceRegistry->get($resource->getAlias());
 
+        if (null === $operation->getTemplate()) {
+            $templateDir = $resource->getTemplatesDir() ?? '';
+            $template = sprintf('%s/%s.html.twig', $templateDir, $operation->getShortName() ?? '');
```
lchrusciel, line 127: Do we really want to search for templates in `/create.html.twig` and sometimes even `/.html.twig`? I would suggest we go with this behaviour only if `getTemplatesDir` is set. And if we want to provide some default folder for templates, it should be part of the configuration of ResourceBundle

src/Component/spec/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactorySpec.php
```diff
@@ -335,4 +336,59 @@ function it_creates_resource_metadata_with_form_type(RegistryInterface $resource
         $operation->getRepository()->shouldReturn('app.repository.dummy');
         $operation->getFormType()->shouldReturn('App\Form\DummyType');
     }
+
+    function it_creates_resource_metadata_with_templates_dir(RegistryInterface $resourceRegistry): void
```
lchrusciel, line 340: for now, we are missing specs for config, where the template is null, and both templatedir and/or short name are null

src/Component/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
loic425: default template dir could be the the resource name.
Example: Resource(name: 'book') will have book as default template dir

I don't want any Resource configuration on the bundle itself to configure the resources at the end.

lchrusciel: But then, if you would like to use the default configuration, you are forced to duplicate templates from default folders (as every default folder is different). We need something better. Perhaps a super simple template in ResourceBundle? I'm still struggling with how to solve the issue properly for cases like right now Sylius show template (which does not exists and magically asks you for it)

PR #592, opened 2023-02-03
Twig context factory

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

To bring a bc-layer for current AdminBundle CRUD templates (or Monofony ones) without any changes, we have to add RequestConfiguration and Metadata on Twig.

```php
        if ($configuration->isHtmlRequest()) {
            return $this->render($configuration->getTemplate(ResourceActions::SHOW . '.html'), [
                'configuration' => $configuration,
                'metadata' => $this->metadata,
                'resource' => $resource,
                $this->metadata->getName() => $resource,
            ]);
        }
```

### DIFF
--- a/src/Bundle/Twig/Context/LegacyContextFactory.php
+++ b/src/Bundle/Twig/Context/LegacyContextFactory.php
@@ -1,0 +1,45 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ResourceBundle\Twig\Context;
+
+use Sylius\Bundle\ResourceBundle\Context\Option\RequestConfigurationOption;
+use Sylius\Component\Resource\Context\Context;
+use Sylius\Component\Resource\Context\Option\MetadataOption;
+use Sylius\Component\Resource\Metadata\Operation;
+use Sylius\Component\Resource\Twig\Context\Factory\ContextFactoryInterface;
+
+final class LegacyContextFactory implements ContextFactoryInterface
+{
+    public function __construct(private ContextFactoryInterface $decorated)
+    {
+    }
+
+    public function create(mixed $data, Operation $operation, Context $context): array
+    {
+        $twigContext = $this->decorated->create($data, $operation, $context);
+
+        $requestConfiguration = $context->get(RequestConfigurationOption::class)?->requestConfiguration();
+        $metadata = $context->get(MetadataOption::class)?->metadata();
+
+        if (null !== $requestConfiguration) {
+            $twigContext['configuration'] = $requestConfiguration;
+        }
+
+        if (null !== $metadata) {
+            $twigContext['metadata'] = $metadata;
+        }
+
+        return $twigContext;
+    }
+}
--- a/src/Component/Context/Option/MetadataOption.php
+++ b/src/Component/Context/Option/MetadataOption.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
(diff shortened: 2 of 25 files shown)


src/Component/spec/Twig/Context/Factory/ContextFactorySpec.php
```diff
@@ -0,0 +1,53 @@
...
+namespace spec\Sylius\Component\Resource\Twig\Context\Factory;
+
+use PhpSpec\ObjectBehavior;
+use Sylius\Component\Resource\Context\Context;
+use Sylius\Component\Resource\Metadata\Index;
+use Sylius\Component\Resource\Metadata\Resource;
+use Sylius\Component\Resource\Metadata\Show;
+use Sylius\Component\Resource\Twig\Context\Factory\ContextFactory;
+
+class ContextFactorySpec extends ObjectBehavior
```
lchrusciel:
```suggestion
final class ContextFactorySpec extends ObjectBehavior
```

src/Component/Context/Option/MetadataOption.php
```diff
@@ -0,0 +1,28 @@
...
+
+final class MetadataOption
+{
+    public function __construct(private MetadataInterface $metadata)
+    {
+    }
+
+    public function metadata(): MetadataInterface
+    {
+        return $this->metadata;
```
lchrusciel, line 26: Why did you introduce these wrapping objects?

lchrusciel (commented):
can you explain me briefly what is happening here?

src/Component/Context/Option/MetadataOption.php
loic425: I've added a comment on the PR description itself.

src/Bundle/Twig/Context/LegacyContextFactory.php
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Sylius\Bundle\ResourceBundle\Twig\Context;
+
+use Sylius\Bundle\ResourceBundle\Context\Option\RequestConfigurationOption;
+use Sylius\Component\Resource\Context\Context;
+use Sylius\Component\Resource\Context\Option\MetadataOption;
+use Sylius\Component\Resource\Metadata\Operation;
+use Sylius\Component\Resource\Twig\Context\Factory\ContextFactoryInterface;
+
+final class LegacyContextFactory implements ContextFactoryInterface
```
lchrusciel, line 22: Shouldn't we already deprecate it? How long do we want to keep it? 2.0? 3.0?

src/Component/Context/Option/MetadataOption.php
lchrusciel: I now get the whole point of the PR, but not the reason of this class

lchrusciel: Ok, to do `$context->get(MetadataOption::class)`. Now I get it. Still, not sure if we shouldn't use `MetadataInterface` instead

src/Component/Twig/Context/Factory/ContextFactory.php
```diff
@@ -0,0 +1,44 @@
...
+        } else {
+            $twigContext['resource'] = $data;
+            $name = $operation->getResource()?->getName();
+
+            if (null !== $name) {
+                $twigContext[$name] = $data;
+            }
+        }
+
+        return $twigContext;
```
lchrusciel, line 42:
```suggestion
        if ($operation instanceof CollectionOperationInterface) {
            $twigContext['resources'] = $data;
            $pluralName = $operation->getResource()?->getPluralName();

            if (null !== $pluralName) {
                $twigContext[$pluralName] = $data;
            }
            
            return $twigContext;
        }

        $twigContext['resource'] = $data;
        $name = $operation->getResource()?->getName();

        if (null !== $name) {
            $twigContext[$name] = $data;
        }

        return $twigContext;
```

lchrusciel (approved):

src/Bundle/Twig/Context/LegacyContextFactory.php
loic425: That's a good question, cause it depends on UI changes.
I don't know if this is a good idea to deprecate its usage, cause to remove deprecation in logs, we have to remove definition of this service in DI in customer project.

PR #595, opened 2023-02-07
Add DI configuration for services

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

### DIFF
--- a/src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineODMDriver.php
+++ b/src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineODMDriver.php
@@ -69,6 +69,7 @@ final class DoctrineODMDriver extends AbstractDoctrineDriver
             $unitOfWorkDefinition,
             $this->getClassMetadataDefinition($metadata),
         ]);
+        $definition->addTag('sylius.repository');
 
         $container->setDefinition($metadata->getServiceId('repository'), $definition);
     }
--- a/src/Bundle/Resources/config/services/state.xml
+++ b/src/Bundle/Resources/config/services/state.xml
@@ -1,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+
+ This file is part of the Sylius package.
+
+ (c) Paweł Jędrzejewski
+
+ For the full copyright and license information, please view the LICENSE
+ file that was distributed with this source code.
+
+-->
+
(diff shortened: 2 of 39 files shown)


src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineODMDriver.php
```diff
@@ -69,6 +69,7 @@ protected function addRepository(ContainerBuilder $container, MetadataInterface
             $unitOfWorkDefinition,
             $this->getClassMetadataDefinition($metadata),
         ]);
+        $definition->addTag('sylius.repository');
```
lchrusciel, line 72: Let's keep repository tag for repositories :) 
```suggestion
        $definition->addTag('sylius.doctrine_driver');
```

lchrusciel (approved):

src/Bundle/Resources/config/services/state.xml
```diff
@@ -0,0 +1,54 @@
...
+
+ For the full copyright and license information, please view the LICENSE
+ file that was distributed with this source code.
+
+-->
+
+<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+        <service id="sylius.state.provider" class="Sylius\Component\Resource\State\Provider">
+            <argument type="tagged_locator" tag="sylius.state_provider" />
```
lchrusciel, line 17: I'm afraid, these tagged locators are broken because we needed to define a key that will be taken into account. What is more, I would recommend adding a default indexing method and adding them as an optional method (as a comment above the interface). Did you test them out?

src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineODMDriver.php
loic425: Hum I've checked and this is not the driver definition, but the repository definition 
https://github.com/loic425/SyliusResourceBundle/blob/317afb718a830576ffe11aef602a5cb0aeed3a5e/src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineODMDriver.php#L66

loic425: so app.repository.book will be tagged with "sylius.repository" and that makes sense :D

PR #608, opened 2023-02-20
No driver

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

When configuring a resource which is not managed by Doctrine

image

### DIFF
--- a/src/Bundle/DependencyInjection/Driver/DriverProvider.php
+++ b/src/Bundle/DependencyInjection/Driver/DriverProvider.php
@@ -37,6 +37,8 @@ final class DriverProvider
             return self::$drivers[$type];
         }
 
+        Assert::notFalse($type, sprintf('No driver was configured on the resource "%s".', $metadata->getAlias()));
+
         switch ($type) {
             case SyliusResourceBundle::DRIVER_DOCTRINE_ORM:
                 return self::$drivers[$type] = new DoctrineORMDriver();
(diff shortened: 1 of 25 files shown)


src/Bundle/DependencyInjection/Driver/DriverProvider.php
```diff
@@ -45,6 +46,8 @@ public static function get(MetadataInterface $metadata): DriverInterface
                 return self::$drivers[$type] = new DoctrinePHPCRDriver();
         }
 
+        Assert::notFalse($type, sprintf('No driver was configured on the resource "%s".', $metadata->getAlias()));
```
lchrusciel: Why do we check it after switch?

src/Bundle/test/config/sylius/resources.yaml
```diff
@@ -3,6 +3,11 @@ sylius_resource:
         locale_provider: test.translation_locale_provider
 
     resources:
+        app.board_game:
+            driver: false
```
lchrusciel, line 7: Why don't we use `null` here?

loic425: null is for default value and this will be configured with doctrine/orm value instead.

lchrusciel: But if we have a default value if null will be provided, this false needs to be defined explicitly. And then, we trigger an error. I don't see the use case for this functionality. Can you write some examples of usage of it?

loic425: It does not trigger an error
https://github.com/Sylius/SyliusResourceBundle/blob/1fcf7dcac28f26217736cc6febd28c550859d3d3/src/Bundle/DependencyInjection/SyliusResourceExtension.php#L145

lchrusciel (approved):

PR #611, opened 2023-02-20
Bulk delete

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Bundle/test/src/Subscription/Entity/Subscription.php
+++ b/src/Bundle/test/src/Subscription/Entity/Subscription.php
@@ -16,6 +16,7 @@ namespace App\Subscription\Entity;
 use App\Subscription\Form\Type\SubscriptionType;
 use Doctrine\ORM\Mapping as ORM;
 use Sylius\Component\Resource\Metadata\ApplyStateMachineTransition;
+use Sylius\Component\Resource\Metadata\BulkDelete;
 use Sylius\Component\Resource\Metadata\Create;
 use Sylius\Component\Resource\Metadata\Delete;
 use Sylius\Component\Resource\Metadata\Index;
@@ -23,7 +24,6 @@ use Sylius\Component\Resource\Metadata\Resource;
 use Sylius\Component\Resource\Metadata\Show;
 use Sylius\Component\Resource\Metadata\Update;
 use Sylius\Component\Resource\Model\ResourceInterface;
-use Symfony\Component\Uid\Uuid;
 use Symfony\Component\Validator\Constraints as Assert;
 
 #[Resource(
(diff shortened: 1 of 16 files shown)


src/Bundle/test/src/Subscription/Entity/Subscription.php
```diff
@@ -33,25 +33,25 @@
...
+        public ?int $id = null,
         #[Assert\NotBlank]
         #[Assert\Email]
         #[ORM\Column(name: 'name', nullable: false)]
         public ?string $email = null,
     ) {
     }
 
-    public function getId(): ?Uuid
+    public function getId(): ?int
```
loic425, line 63: I've simplified this cause I had some troubles. But it works on a test project with an uuid on a mysql database.

lchrusciel: What were the troubles?

src/Bundle/test/src/Tests/Controller/SubscriptionUiTest.php
```diff
@@ -99,7 +99,7 @@ public function it_allows_updating_a_subscription(): void
             'sylius_resource[email]' => 'biff.tannen@bttf.com',
         ]);
 
-        $this->assertResponseRedirects(null, expectedCode: Response::HTTP_FOUND);
+        //$this->assertResponseRedirects(null, expectedCode: Response::HTTP_FOUND);
```
lchrusciel: Is it possible to international?

src/Bundle/test/src/Tests/Controller/SubscriptionUiTest.php
```diff
@@ -137,6 +137,22 @@ public function it_allows_deleting_a_subscription(): void
...
 
+    /** @test */
+    public function it_allows_deleting_multiple_subscriptions(): void
+    {
+        $this->loadFixturesFromFile('subscriptions.yml');
+
+        $this->client->request('GET', '/admin/subscriptions');
+        $this->client->submitForm('Bulk delete');
+
+        $this->assertResponseRedirects(null, expectedCode: Response::HTTP_FOUND);
```
lchrusciel, line 148: Is this test correct?

src/Bundle/test/src/Subscription/Entity/Subscription.php
loic425: The findById($ids) did not work with SQLite, but I have to check it that again later.
But if there is a bug with that on SQLite, we already have a bug on current implementation.

src/Bundle/test/src/Tests/Controller/SubscriptionUiTest.php
loic425: That was a mistake, the redirection is OK.

src/Bundle/test/src/Tests/Controller/SubscriptionUiTest.php
loic425: Yes, It checks that the redirect is ok. (it's a code from @Zales0123 on ScienceBookUiTest)

lchrusciel (approved):

PR #612, opened 2023-02-21
Add state machine processor, listener & operation

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Bundle/test/config/packages/test/sylius_grid.yaml
+++ b/src/Bundle/test/config/packages/test/sylius_grid.yaml
@@ -1,6 +1,7 @@
 sylius_grid:
     templates:
         action:
+            apply_transition: "grid/action/apply_transition.html.twig"
             delete: 'grid/action/delete.html.twig'
             show: 'grid/action/show.html.twig'
             update: 'grid/action/update.html.twig'
--- a/src/Bundle/test/config/packages/test/sylius_grid.yaml
+++ b/src/Bundle/test/config/packages/test/sylius_grid.yaml
@@ -1,7 +1,7 @@
 sylius_grid:
     templates:
         action:
-            apply_transition: "grid/action/apply_transition.html.twig"
+            apply_transition: 'grid/action/apply_transition.html.twig'
             delete: 'grid/action/delete.html.twig'
             show: 'grid/action/show.html.twig'
             update: 'grid/action/update.html.twig'
--- a/src/Bundle/test/config/packages/test/winzou_state_machine.yaml
+++ b/src/Bundle/test/config/packages/test/winzou_state_machine.yaml
@@ -32,3 +32,18 @@ winzou_state_machine:
             reopen:
                 from: [ closed ]
                 to: review
+    subscription:
+        class: App\Subscription\Entity\Subscription
+        property_path: state
+        graph: subscription
+        states:
+            - new
+            - accepted
+            - rejected
+        transitions:
+            accept:
+                from: [ new ]
+                to: accepted
+            reject:
+                from: [ new ]
+                to: rejected
--- a/src/Bundle/test/src/Subscription/Entity/Subscription.php
+++ b/src/Bundle/test/src/Subscription/Entity/Subscription.php
@@ -65,12 +65,12 @@ class Subscription implements ResourceInterface
         return $this->id;
     }
 
-    public function getCurrentState(): string
+    public function getState(): string
     {
         return $this->state;
     }
 
-    public function setCurrentState(string $currentState): void
+    public function setState(string $currentState): void
     {
         $this->state = $currentState;
     }
--- a/src/Bundle/test/src/Subscription/Entity/Subscription.php
+++ b/src/Bundle/test/src/Subscription/Entity/Subscription.php
@@ -13,7 +13,9 @@ declare(strict_types=1);
 
 namespace App\Subscription\Entity;
 
+use App\Subscription\Form\Type\SubscriptionType;
 use Doctrine\ORM\Mapping as ORM;
+use Sylius\Component\Resource\Metadata\ApplyStateMachineTransition;
 use Sylius\Component\Resource\Metadata\Create;
 use Sylius\Component\Resource\Metadata\Delete;
 use Sylius\Component\Resource\Metadata\Index;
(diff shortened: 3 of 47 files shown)


src/Component/Metadata/Index.php
```diff
@@ -36,6 +36,9 @@ public function __construct(
         ?bool $write = null,
         ?string $formType = null,
         ?array $formOptions = null,
+        ?string $stateMachineComponent = null,
+        ?string $stateMachineTransition = null,
+        ?string $stateMachineGraph = null,
```
lchrusciel: I'm quite sure, we don't want to support state machine transition on Index operation

src/Component/Metadata/HttpOperation.php
```diff
@@ -34,8 +34,12 @@ public function __construct(
         ?string $grid = null,
         ?bool $read = null,
         ?bool $write = null,
+        ?bool $validate = null,
         ?string $formType = null,
         ?array $formOptions = null,
+        ?string $stateMachineComponent = null,
+        ?string $stateMachineTransition = null,
+        ?string $stateMachineGraph = null,
```
lchrusciel: This should not be generic operation in my opinion. Only creating and updating should take advantage of it

src/Component/Metadata/Delete.php
```diff
@@ -35,6 +35,9 @@ public function __construct(
         ?bool $write = null,
         ?string $formType = null,
         ?array $formOptions = null,
+        ?string $stateMachineComponent = null,
+        ?string $stateMachineTransition = null,
+        ?string $stateMachineGraph = null,
```
lchrusciel: I'm wondering if we support state machine transition on delete. Can you check?

src/Component/Metadata/Show.php
```diff
@@ -36,6 +36,9 @@ public function __construct(
         ?bool $write = null,
         ?string $formType = null,
         ?array $formOptions = null,
+        ?string $stateMachineComponent = null,
+        ?string $stateMachineTransition = null,
+        ?string $stateMachineGraph = null,
```
lchrusciel:
```suggestion
```

src/Bundle/test/src/Subscription/Entity/Subscription.php
```diff
@@ -27,23 +29,30 @@
...
     section: 'admin',
+    formType: SubscriptionType::class,
     templatesDir: 'crud',
     routePrefix: '/admin',
 )]
 #[Index(grid: 'app_subscription')]
 #[Create]
 #[Update]
+#[ApplyStateMachineTransition(stateMachineTransition: 'accept')]
+#[ApplyStateMachineTransition(stateMachineTransition: 'reject')]
```
lchrusciel, line 40: Is it enough data to add such an operation? Don't we miss the graph here? How does it work?

src/Component/Metadata/Delete.php
loic425: I see no state machine transition usage on deleteAction, but I only see state machine usage on createAction but not on updateAction and that's weird.

src/Bundle/test/src/Subscription/Entity/Subscription.php
loic425: That graph is an optional value. It's only when you use several state machines on a same resource.

src/Component/Metadata/HttpOperation.php
loic425: Hum indeed, so maybe with an interface to know if state machine is applicable.

src/Bundle/test/src/Subscription/Entity/Subscription.php
lchrusciel: How do you fetch it, then? Do we expect it to be called default, or do we just fetch the only existing for given class?

src/Bundle/test/config/packages/test/sylius_grid.yaml
```diff
@@ -1,6 +1,7 @@
 sylius_grid:
     templates:
         action:
+            apply_transition: "grid/action/apply_transition.html.twig"
```
lchrusciel:
```suggestion
            apply_transition: 'grid/action/apply_transition.html.twig'
```

src/Bundle/test/config/packages/test/winzou_state_machine.yaml
```diff
@@ -32,3 +32,18 @@ winzou_state_machine:
             reopen:
                 from: [ closed ]
                 to: review
+    subscription:
+        class: App\Subscription\Entity\Subscription
+        property_path: state
```
lchrusciel, line 37: Minor thing,  but we are mixing `state` and `currentState` variables. Property is called state, but getter/setter are for `currentState`. TBH, I'm surprised this configuration even work, I would expect exception here, as there is no `setState` method on Subscription

src/Bundle/test/src/Tests/Controller/SubscriptionUiTest.php
```diff
@@ -137,6 +137,23 @@ public function it_allows_deleting_a_subscription(): void
         $this->assertEmpty($subscriptions);
     }
 
+    /** @test */
+    public function it_allows_accepting_a_subscription(): void
```
lchrusciel, line 141: Is this test executed with Symfony state machine? Do we have the same for winzou?

src/Component/Metadata/MetadataInterface.php
```diff
@@ -13,6 +13,9 @@
 
 namespace Sylius\Component\Resource\Metadata;
 
+/**
+ * @method string|null getStateMachineComponent()
```
lchrusciel, line 17: Do we need this soft method requirement?

src/Bundle/test/src/Subscription/Entity/Subscription.php
loic425: On Winzou, the graph is called "default" is you do not specified it =>
 `$graph = $operation->getStateMachineGraph() ?? 'default';`

This is here https://github.com/Sylius/SyliusResourceBundle/pull/612/files#diff-fccfca1acf785d9965e1c335fbcc071e6dffb7433489d6d2341b760ae68e2b4dR36

On Workflow, this is the "workflow name" and is value is null by default
https://github.com/Sylius/SyliusResourceBundle/pull/612/files#diff-52c7ef7ce600296cd01a16a35c3c604f6bc0ab7e8b8179138ad17111cf912267R36

src/Bundle/test/config/packages/test/winzou_state_machine.yaml
loic425: Indeed, I think this works cause my property is public. I'm gonna try to change that.

src/Bundle/test/src/Tests/Controller/SubscriptionUiTest.php
loic425: I think so, cause we run with the two components on the CI and we have no "specific state machine configuration" for this resource as we can do now.

src/Component/Metadata/MetadataInterface.php
loic425: Yeah I think that's better to inform that this method has been introduced.

loic425: Hum that let me think that I have to check if this method exists when I use it on a service.
Hum.. What do you think?

src/Bundle/test/src/Subscription/Entity/Subscription.php
lchrusciel: Exactly, but the name of the graph in winzou is `graph: subscription`. Therefore, it shouldn't work out-of-the-box. I assume that winzou config is not tested. Am I correct?

src/Component/Metadata/MetadataInterface.php
lchrusciel: either new interface or always expect this method to be present

lchrusciel (approved):
Let's merge it, but also add task for refactor of winzou tests to confirm, that it works as expected

PR #616, opened 2023-02-26
Allow doctrine/collections ^2.0

### LINKED ISSUE #494: Add support for doctrine/collections 2.x
It requires PHP 8.1.

dannyvw:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no 
| Related tickets | fixes #494 
| License         | MIT

Not sure if this is right way to fix the CI

When installing doctrine/collections ^2.0 doctrine/persistence ^3.0 and PHP 8.1 should be used.
doctrine/collections ^2.0 should not be installed on PHP 8.0

What to do with the psalm errors?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ on:
 jobs:
     tests:
         runs-on: ubuntu-latest
-        name: "PHP ${{ matrix.php }}, Symfony ${{ matrix.symfony }}, Twig ${{ matrix.twig }}, Persistence: ${{ matrix.persistence }}"
+        name: "PHP ${{ matrix.php }}, Symfony ${{ matrix.symfony }}, Twig ${{ matrix.twig }}, Persistence: ${{ matrix.persistence }}, Collections: ${{ matrix.collections }}"
         strategy:
             fail-fast: false
             matrix:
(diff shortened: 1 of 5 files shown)


loic425:
@dannyvw Could you rebase that PR please?

dannyvw:
@loic425 Rebased but still a few errors. I have currently no time to fix this. Maybe you can have a look?

dannyvw:
@loic425 I have pushed some changes.

.github/workflows/build.yml
```diff
@@ -22,17 +22,27 @@ jobs:
...
-                        symfony: "~6.2.0"
+                    - php: "8.0"
+                      symfony: "~6.1.0"
+                    - php: "8.0"
+                      symfony: "~6.2.0"
+                    - php: "8.0"
+                      collections: "^2.0"
+                    - php: "8.0"
+                      collections: "^2.0"
+                      persistence: "^2.0"
```
oallain: IMO this case is covered by the above exclusion.

dannyvw:
@loic425 Could you check this please?

oallain (approved):

loic425 (approved):

src/Bundle/Form/DataTransformer/RecursiveTransformer.php
```diff
@@ -28,7 +29,7 @@ public function __construct(DataTransformerInterface $decoratedTransformer)
     }
 
     /** @param Collection|null $value */
-    public function transform($value): Collection
+    public function transform($value): ReadableCollection
```
Zales0123, line 32: I had mixed feelings about whether we're not breaking anything here - in the end, it's a change of the type of returned value. But because it's a final class, used only inside forms processing and in fact, the BC is not broken, as we could just return _more_ not _less_... It's safe to have it merged 🖖

Zales0123 (approved):

Zales0123:
Thank you, @dannyvw! 🎉

src/Bundle/Form/DataTransformer/RecursiveTransformer.php
loic425: Yes, I've checked it (cf our slack), there's no BC break

PR #617, opened 2023-02-27
Ajax support (json)

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Bundle/Resources/config/services/listener.xml
+++ b/src/Bundle/Resources/config/services/listener.xml
@@ -13,11 +13,25 @@
 
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
+        <service id="sylius.negotiator" class="Negotiation\Negotiator" />
+
+        <service id="sylius.listener.add_format" class="Sylius\Component\Resource\Symfony\EventListener\AddFormatListener">
+            <argument type="service" id="sylius.resource_metadata_operation.initiator.http_operation" />
+            <argument type="service" id="sylius.negotiator" />
+            <tag name="kernel.event_listener" event="kernel.request" priority="28" />
+        </service>
+
         <service id="sylius.listener.read" class="Sylius\Component\Resource\Symfony\EventListener\ReadListener">
             <argument type="service" id="sylius.resource_metadata_operation.initiator.http_operation" />
             <argument type="service" id="sylius.context.initiator.request_context" />
             <argument type="service" id="sylius.state.provider" />
             <tag name="kernel.event_listener" event="kernel.request" priority="4" />
+        </service>
+
+        <service id="sylius.listener.deserialize" class="Sylius\Component\Resource\Symfony\EventListener\DeserializeListener">
+            <argument type="service" id="sylius.resource_metadata_operation.initiator.http_operation" />
+            <argument type="service" id="serializer" on-invalid="null" />
+            <tag name="kernel.event_listener" event="kernel.request" priority="2" />
         </service>
 
         <service id="sylius.listener.form" class="Sylius\Component\Resource\Symfony\EventListener\FormListener">
@@ -29,6 +43,7 @@
 
         <service id="sylius.listener.validate" class="Sylius\Component\Resource\Symfony\EventListener\ValidateListener">
             <argument type="service" id="sylius.resource_metadata_operation.initiator.http_operation" />
+            <argument type="service" id="validator" />
             <tag name="kernel.event_listener" event="kernel.view" priority="64" />
         </service>
 
@@ -37,6 +52,12 @@
             <argument type="service" id="sylius.context.initiator.request_context" />
             <argument type="service" id="sylius.state.processor" />
             <tag name="kernel.event_listener" event="kernel.view" priority="32" />
+        </service>
+
+        <service id="sylius.listener.serialize" class="Sylius\Component\Resource\Symfony\EventListener\SerializeListener">
+            <argument type="service" id="sylius.resource_metadata_operation.initiator.http_operation" />
+            <argument type="service" id="serializer" on-invalid="null" />
+            <tag name="kernel.event_listener" event="kernel.view" priority="16" />
         </service>
 
         <service id="sylius.listener.flash" class="Sylius\Component\Resource\Symfony\EventListener\FlashListener">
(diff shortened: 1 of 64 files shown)


src/Bundle/Routing/OperationRouteFactory.php
```diff
@@ -59,32 +63,55 @@ private function getDefaultRoutePathForOperation(MetadataInterface $metadata, Ht
             throw new \InvalidArgumentException(sprintf('Operation "%s" should have a short name. Please define one.', $operation::class));
         }
 
-        if ('index' === $shortName) {
-            return sprintf('%s', $rootPath);
+        if ($operation instanceof CollectionOperationInterface) {
```
lchrusciel, line 68: This file seems to be a great candidate to split into several smaller ones

src/Bundle/test/src/Tests/Controller/XmlApiTestCase.php
```diff
@@ -0,0 +1,30 @@
...
+declare(strict_types=1);
+
+namespace App\Tests\Controller;
+
+use ApiTestCase\XmlApiTestCase as BaseXmlApiTestCase;
+use Symfony\Component\HttpFoundation\Response;
+
+abstract class XmlApiTestCase extends BaseXmlApiTestCase
+{
+    protected static array $headersWithContentType = [
```
lchrusciel, line 21: Should we open a PR with it to ApiTestCase?

src/Component/Symfony/EventListener/AddFormatListener.php
```diff
@@ -0,0 +1,63 @@
...
+
+        /** @var string|null $routeFormat */
+        $routeFormat = $request->attributes->get('_format') ?? null;
+
+        if (null === $routeFormat) {
+            $mimeTypes = ['application/json', 'application/xml'];
+        } else {
+            $mimeTypes = Request::getMimeTypes($routeFormat);
+        }
+
```
lchrusciel:
```suggestion
        /** @var string|null $routeFormat */
        $routeFormat = $request->attributes->has('_format') ? Request::getMimeTypes($request->attributes->get('_format')) : ['application/json', 'application/xml'];
```

src/Component/Symfony/EventListener/AddFormatListener.php
```diff
@@ -0,0 +1,63 @@
...
+        $accept = $request->headers->get('Accept');
+
+        if (null === $accept) {
+            return;
+        }
```
lchrusciel: According to these line, there is no point in resolving mime types, if we don't have accept header. Let's switch the order then - firstly, let's check for accept header, secondly let's try to resolve mime types.

src/Component/Symfony/EventListener/AddFormatListener.php
```diff
@@ -0,0 +1,63 @@
...
+
+        /** @var BaseAccept|null $mediaType */
+        $mediaType = $this->negotiator->getBest($accept, $mimeTypes);
+
+        if (null === $mediaType) {
+            return;
+        }
+
+        $request->setRequestFormat($request->getFormat($mediaType->getType()));
```
lchrusciel: Why do we have to set request format on our own? Why can't we trust Symfony on that?

src/Bundle/Resources/config/services/listener.xml
```diff
@@ -52,5 +73,10 @@
             <argument type="service" id="sylius.state.responder" />
             <tag name="kernel.event_listener" event="kernel.view" priority="8" />
         </service>
+
+        <service id="sylius.listener.exception.validation" class="Sylius\Component\Resource\Symfony\Validator\EventListener\ValidationExceptionListener">
+            <argument type="service" id="serializer" />
```
lchrusciel: why this dependency is not optional?
```suggestion

            <argument type="service" id="serializer" on-invalid="null" />
```

src/Component/Symfony/Request/State/ApiResponder.php
```diff
@@ -0,0 +1,65 @@
...
+
+        /** @var string $mimeType */
+        $mimeType = $request->getMimeType($format);
+
+        $headers = [
+            'Content-Type' => sprintf('%s; charset=utf-8', $mimeType),
+            'Vary' => 'Accept',
+            'X-Content-Type-Options' => 'nosniff',
+            'X-Frame-Options' => 'deny',
+        ];
```
lchrusciel, line 51: How much should we enforce these headers? Is there a value to extract it to separate service?

src/Component/Symfony/Request/State/Responder.php
```diff
@@ -0,0 +1,45 @@
...
+use Sylius\Component\Resource\Context\Option\RequestOption;
+use Sylius\Component\Resource\Metadata\Operation;
+use Sylius\Component\Resource\State\ResponderInterface;
+
+final class Responder implements ResponderInterface
+{
+    public function __construct(
+        private ResponderInterface $htmlResponder,
+        private ResponderInterface $apiResponder,
+    ) {
```
lchrusciel, line 26: As we have to identical responders, it may happen that they will switched in places. What would you say about usage of locator instead? One with `html` tag and second one with defined `api` keys

src/Component/Symfony/Request/State/TwigResponder.php
```diff
@@ -51,7 +51,6 @@ public function respond(mixed $data, Operation $operation, Context $context): ?R
 
         if ($operation instanceof DeleteOperationInterface && $operation instanceof HttpOperation) {
             return $this->redirectHandler->redirectToResource($data, $operation, $request);
-            //return $this->redirectHandler->redirectToIndex($data, $operation, $request);
```
lchrusciel, line 54: Should be removed?

src/Bundle/test/src/Tests/Controller/XmlApiTestCase.php
loic425: good idea, cause Symfony will return a **text/xml** as response

src/Component/Symfony/Request/State/ApiResponder.php
loic425: Good point, as you can guess, theses headers are hardcoded in APIPlatform core.

@dunglas what do you think?

src/Component/Symfony/Request/State/Responder.php
loic425: Hum the idea was to let the developers use its own html/api responder.

src/Component/Symfony/Request/State/TwigResponder.php
loic425: Yes I think I did in a future PR lol

src/Bundle/Routing/OperationRouteFactory.php
loic425: Yeah, I also have to move this routing factory into the component (an issue was created on the board).
We can address an issue on the board for this one?

lchrusciel (approved):

lchrusciel:
Thank you, @loic425!

PR #619, opened 2023-03-01
Add normalization & denormalization contexts

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | f
| License         | MIT


src/Bundle/test/src/Subscription/Entity/Subscription.php
```diff
@@ -47,6 +48,8 @@
     alias: 'app.subscription',
     section: 'ajax',
     routePrefix: '/ajax',
+    normalizationContext: ['groups' => 'subscription:read'],
```
lchrusciel, line 51: A small thing that we could improve here is the usage of PHP consts as groups. Not needed, but recently I started to think about it as the default approach and wanted to share this thought :)

lchrusciel (approved):

lchrusciel:
Thank you, @loic425!

PR #624, opened 2023-03-06
[Tests] Add test for Grid with custom repository method

### LINKED ISSUE #626: Checking grid with repository methods & arguments

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | Fixes #626 
| License         | MIT

### DIFF
--- a/src/Bundle/test/src/Subscription/Entity/Subscription.php
+++ b/src/Bundle/test/src/Subscription/Entity/Subscription.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace App\Subscription\Entity;
 
 use App\Subscription\Form\Type\SubscriptionType;
+use App\Subscription\Repository\SubscriptionRepository;
 use Doctrine\ORM\Mapping as ORM;
 use Sylius\Component\Resource\Metadata\ApplyStateMachineTransition;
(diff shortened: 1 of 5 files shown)


src/Bundle/test/src/Repository/BookRepository.php
```diff
@@ -26,4 +27,12 @@ public function __construct(ManagerRegistry $registry)
...
         parent::__construct($registry, Book::class);
     }
+
+    public function createListQueryBuilder(): QueryBuilder
+    {
+        $queryBuilder = $this->createQueryBuilder('o');
+        $queryBuilder->andWhere($queryBuilder->expr()->isNotNull('o.archivedAt'));
+
+        return $queryBuilder;
+    }
```
lchrusciel, line 37: Why did you add this code?

loic425: The goal of this PR is to test if the grid configuration is applied (such as a repository method defined on the grid)

lchrusciel: ok, but where is it used?

PR #644, opened 2023-03-21
Add pre & post operation events

### LINKED ISSUE #643: Pre & Post Listener

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #643
| License         | MIT


src/Bundle/Event/ResourceControllerEvent.php
```diff
@@ -13,103 +13,12 @@
...
-
-    private int $errorCode = 500;
-
-    private ?Response $response = null;
-
-    /**
-     * @psalm-suppress MissingReturnType
-     */
-    public function stop(string $message, string $type = self::TYPE_ERROR, array $parameters = [], int $errorCode = 500)
+if (false) {
```
vvasiloi, line 18: Is this intended?

loic425: yes, this is a bc-layer https://www.schmengler-se.de/en/2016/09/php-using-class_alias-to-maintain-bc-while-move-rename-classes/

vvasiloi: I understand that, but shouldn't it be `if (!class_exists()) { class_alias(); }`?

src/Bundle/Resources/config/services/listener.xml
```diff
@@ -13,13 +13,27 @@
 
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
+        <service id="sylius.negotiator" class="Negotiation\Negotiator" />
```
lchrusciel: Shouldn't we put this class as a direct dependency?

src/Bundle/Event/ResourceControllerEvent.php
lchrusciel: From what I remember, this is how doctrine and api platform is handling class movement

src/Bundle/test/vendor/autoload.php
```diff
@@ -0,0 +1,25 @@
+<?php
```
lchrusciel: Why did you added this class?

src/Bundle/test/src/Subscription/EventSubscriber/SmokeSubscriptionEventsSubscriber.php
```diff
@@ -0,0 +1,58 @@
...
+
+namespace App\Subscription\EventSubscriber;
+
+use Sylius\Bundle\ResourceBundle\Event\ResourceControllerEvent;
+use Sylius\Component\Resource\Symfony\EventDispatcher\GenericEvent;
+use Sylius\Component\Resource\Symfony\EventDispatcher\OperationEvent;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\EventDispatcher\GenericEvent as SymfonyGenericEvent;
+
+class SmokeSubscriptionEventsSubscriber implements EventSubscriberInterface
```
lchrusciel: Missing final

src/Bundle/test/vendor/composer/LICENSE
```diff
@@ -0,0 +1,21 @@
+
```
lchrusciel: Not needed?

src/Bundle/test/vendor/composer/autoload_classmap.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+// autoload_classmap.php @generated by Composer
+
+$vendorDir = dirname(__DIR__);
```
lchrusciel: To remove

src/Bundle/test/vendor/composer/autoload_namespaces.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+// autoload_namespaces.php @generated by Composer
```
lchrusciel: To remove

src/Bundle/test/vendor/composer/autoload_psr4.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+// autoload_psr4.php @generated by Composer
```
lchrusciel: To remove

src/Bundle/test/vendor/composer/autoload_real.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+// autoload_real.php @generated by Composer
```
lchrusciel: To remove

src/Bundle/test/vendor/composer/autoload_static.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+// autoload_static.php @generated by Composer
```
lchrusciel: To remove

src/Component/State/EventDispatcherProcessor.php
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function process(mixed $data, Operation $operation, Context $context): mixed
+    {
+        $this->operationEventDispatcher->dispatchPreEvent($data, $operation, $context);
+
+        $result = $this->decorated->process($data, $operation, $context);
```
lchrusciel, line 38: I wonder if we shouldn't provide an event after processing but before flushing to DB. It would be much closer to what we have right now in Sylius. But that would be something in addition to what we have right now.

src/Component/Symfony/EventDispatcher/OperationEventDispatcher.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        $this->dispatchEvent($data, $operation, $context, 'post');
+    }
+
+    public function dispatchInitializeEvent(mixed $data, Operation $operation, Context $context): void
+    {
+        $this->dispatchEvent($data, $operation, $context, 'initialize');
+    }
+
+    private function dispatchEvent(mixed $data, Operation $operation, Context $context, ?string $eventType = null): void
```
lchrusciel, line 52: Could you extract this logic to the separate class?

src/Component/State/BulkAwareProcessor.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        if (
+            !$operation instanceof BulkOperationInterface ||
+            !\is_iterable($data)
+        ) {
+            return $this->decorated->process($data, $operation, $context);
+        }
+
+        foreach ($data as $item) {
+            $this->decorated->process($item, $operation, $context);
```
lchrusciel, line 43: Can you confirm, that for bulk operations we are still triggering pre and post events for each item in bulk? Also, when do we flush in such situation

src/Component/spec/State/ProcessorSpec.php
```diff
@@ -19,13 +19,14 @@
 use Sylius\Component\Resource\Metadata\Create;
 use Sylius\Component\Resource\State\Processor;
 use Sylius\Component\Resource\State\ProcessorInterface;
+use Sylius\Component\Resource\Symfony\EventDispatcher\OperationEventDispatcherInterface;
 use Sylius\Component\Resource\Tests\Dummy\ProcessorWithCallable;
 
 final class ProcessorSpec extends ObjectBehavior
 {
-    function let(ContainerInterface $locator): void
+    function let(ContainerInterface $locator, OperationEventDispatcherInterface $operationEventDispatcher): void
```
lchrusciel: Should be reverted

src/Bundle/Event/ResourceControllerEvent.php
loic425: Yes Lukasz, it is :)

src/Bundle/Resources/config/services/listener.xml
loic425: There is no service for this class. I've just copied/paste what APIP did.

src/Bundle/test/vendor/autoload.php
loic425: This is a mistake. I've run a "composer autoload".

src/Bundle/test/vendor/composer/LICENSE
loic425: yep, not needed

src/Component/State/EventDispatcherProcessor.php
loic425: The decorated one is the Doctrine one when there is a Doctrine driver configured.

src/Component/Symfony/EventDispatcher/OperationEventDispatcher.php
loic425: to an event factory one?

src/Bundle/test/src/Subscription/EventSubscriber/SmokeSubscriptionEventsSubscriber.php
loic425: done

src/Bundle/test/vendor/composer/LICENSE
loic425: removed

src/Bundle/test/vendor/composer/autoload_static.php
loic425: removed

src/Component/State/BulkAwareProcessor.php
loic425: Yes, we trigger a pre and post events for each item. There is a EventDispatcherBulkAwareProcessor with do that.
I've checked and everything seems ok, but we can check together before releasing the alpha. Maybe we can add some logic into the "Smoke subscriber" to log that and check if everything is ok.

src/Component/spec/State/ProcessorSpec.php
loic425: done

src/Component/Symfony/EventDispatcher/OperationEventDispatcher.php
lchrusciel: yes. Something like this class could be `OperationsEventDispatcher` and then would have `OperationEventDispatcher` with this logic. Or just make this method public, so others can just simply use it if the customer event type is needed.

lchrusciel:
Thank you, @loic425!

PR #654, opened 2023-03-24
[Performance] Add cached resource metadata collection factory

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

Capture d’écran du 2023-03-24 10-26-24

### DIFF
--- a/src/Bundle/Resources/config/services/metadata.xml
+++ b/src/Bundle/Resources/config/services/metadata.xml
@@ -13,6 +13,10 @@
 
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
+        <service id="sylius.cache.metadata.resource_collection" parent="cache.system" public="false">
+            <tag name="cache.pool" />
+        </service>
+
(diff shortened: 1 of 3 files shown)


src/Bundle/Resources/config/services/metadata.xml
```diff
@@ -13,16 +13,30 @@
 
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
+        <service id="sylius.cache.metadata.resource_collection" parent="cache.system" public="false">
```
loic425, line 16: @lchrusciel 
the services in this file are private by default, but I think it works differently with parent, cause this one is public if I do not tag it as private.

loic425:
@lchrusciel After this PR, it would be interesting to do some Blackfire sessions :dancers:

lchrusciel (approved):

lchrusciel:
But we should do it before and after 😂

lchrusciel:
Thank you, @loic425!

loic425:
> But we should do it before and after 😂

Yes but I'd like to compare resource controller vs operations performances too.

PR #657, opened 2023-03-24
[New documentation] Create a new resource

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes kind of
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

[Preview](https://github.com/Sylius/SyliusResourceBundle/blob/cc11af67b4c111007ab5f2fd3b4235c65b20c983/docs/create_new_resource.md)

### DIFF
--- a/docs/create_new_resource.md
+++ b/docs/create_new_resource.md
@@ -1,6 +1,6 @@
 # Create a new resource
 
-As an example, we create a Book entity with 4 fields:
+As an example, let's create a Book entity:
 
 <!-- TOC -->
 * [Create the entity](#create-the-entity)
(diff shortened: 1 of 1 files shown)


lchrusciel (approved):
This document assumes, in my opinion, that we know what the resource is, why we want to create them, and that is the consequence of having them. The last part is the most needed one, in my opinion. How to check that we succeed? What does it give me?

lchrusciel:
Thank you, @loic425!

docs/create_new_resource.md
```diff
@@ -0,0 +1,164 @@
...
+{
+    use ResourceRepositoryTrait;
+
+    public function __construct(ManagerRegistry $registry)
+    {
+        parent::__construct($registry, Book::class);
+    }
+}
+```
+
```
lchrusciel, line 163: Can we add link to the next documentation page?

docs/create_new_resource.md
```diff
@@ -0,0 +1,164 @@
...
+
+       
+Success! 
+       
+
+Next: When you're ready, create a migration with php bin/console make:migration
+
+```
+
```
lchrusciel, line 99: We should show the result of these commands + we are missing resource tag adding

lchrusciel: Also, don't we need to add some configuration to make it work?

PR #695, opened 2023-04-07
Alias the workflow registry

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes (will be)
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

See these different versions
[5.4](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Bundle/FrameworkBundle/Resources/config/workflow.php)
[6.1](https://github.com/symfony/symfony/blob/6.1/src/Symfony/Bundle/FrameworkBundle/Resources/config/workflow.php)
[6.2](https://github.com/symfony/symfony/blob/6.2/src/Symfony/Bundle/FrameworkBundle/Resources/config/workflow.php)

On 6.2, the 'workflow.registry' is an alias of '.workflow.registry' instead of being the id itself.

### DIFF
--- a/src/Bundle/DependencyInjection/Compiler/RegisterStateMachinePass.php
+++ b/src/Bundle/DependencyInjection/Compiler/RegisterStateMachinePass.php
@@ -106,9 +106,26 @@ final class RegisterStateMachinePass implements CompilerPassInterface
             return;
         }
 
+        $this->defineSyliusWorkflowRegistry($container);
+
         $stateMachineDefinition = $container->register('sylius.resource_controller.state_machine.symfony', Workflow::class);
         $stateMachineDefinition->setPublic(false);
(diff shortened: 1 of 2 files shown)


vvasiloi:
Is there a way to not use the registry and still provide a good DX?

vvasiloi:
After discussing with @loic425 in private we came to the conclusion that we should try to avoid usage of the registry.
The alias was deprecated and the workflow registry is considered an internal service and it's usage is discouraged by the Symfony team.
Sources:
- https://github.com/symfony/symfony/pull/46000
- https://symfony.com/doc/current/workflow.html

image

I'll come up with an alternative proposal to this PR.

PR #708, opened 2023-04-20
Fix undefined kernel.project_dir for mapping paths

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

The issue appeared since we added auto registering resources feature cause we retrieve the mapping path into SyliusResourceExtension. Kernel parameters are not available on Extensions.

## My proposal

We configure it into our project. So we can add this configuration via Flex (EDIT: already merged in https://github.com/symfony/recipes-contrib/pull/1499, so it seems to be the best choice now).
I've added a note to the UPGRADE.md cause it will breaks routes using SyliusCrudRoutes and SyliusRoute attributes.

## Alternative solution (but not great at all):

Add this configuration on prepend config.

This works if we preprend the configuration like this:

```php
$container->prependExtensionConfig('sylius_resource', [
    'mapping' => [
        'paths' => ['%kernel.project_dir%/src/Entity'],
    ],
]);
```

[⚠️](https://emojipedia.org/warning/) But we cannot drop this config easily, if we configure it into the yaml configuration, it will append the new directory or the same directory twice if we add this configuration via Flex.
And there is no Entity directory in the project, it will break. Silent the not found path could be a solution.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,19 @@
+## UPGRADE FOR `1.11.x`
+
+### FROM `1.10.x` to `1.11.x`
+
+We remove the default mapping paths which are used to read PHP 8 attributes to build routes.
+
+To configure this default value, please edit your `config/packages/sylius_resource.yaml` file to add your mapping paths manually:
+
+```yaml
+# config/packages/sylius_resource.yaml
+sylius_resource:
+    mapping:
+        paths:
+            - '%kernel.project_dir%/src/Entity'
+```
+
 ## UPGRADE FOR `1.10.x`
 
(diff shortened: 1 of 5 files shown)


loic425:
The linked Flex recipe https://github.com/symfony/recipes-contrib/pull/1499

vvasiloi (approved):

lchrusciel (approved):

lchrusciel:
Thank you, @loic425!

PR #735, opened 2023-07-07
Drivers can now be configured

Prometee:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

This PR allows to configure any drivers you want or even change the default one.

Here is the previous default configuration :

```yaml
sylius_resource:
    drivers:
        - 'doctrine/orm'
```

Now you can setup a little bit more info :

```yaml
sylius_resource:
    drivers:
        doctrine/orm:
            class: 'Sylius\Bundle\ResourceBundle\DependencyInjection\Driver\Doctrine\DoctrineORMDriver'
```

To keep backward compatibility, you still can set the config with a simple string array when you are using one the 3 available drivers : `doctrine/*`

Right now this PR is using only the `class` config member, but we can have other things added here in the futur.

## Custom driver example

You are creating your own driver and wanted to register it in the list of available drivers, here is how to proceed :

```yaml
# config/packages/sylius_resource.yaml
sylius_resource:
    drivers:
        0: doctrine/orm # Add default value
        custom_driver:
            class: 'MyBundleOrApp\SyliusResource\Driver\MyCustomDriver'
```

```php
<?php

declare(strict_types=1);

namespace MyBundleOrApp\SyliusResource\Driver;

use Sylius\Bundle\ResourceBundle\DependencyInjection\Driver\DriverInterface;
use Sylius\Component\Resource\Metadata\MetadataInterface;
use Symfony\Component\DependencyInjection\ContainerBuilder;

final class MyCustomDriver implement DriverInterface {

    public function getType(): string
    {
        return 'custom_driver';
    }

    public function load(ContainerBuilder $container, MetadataInterface $metadata): void
    {
        ...
    }
}
```

### DIFF
--- a/src/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bundle/DependencyInjection/Configuration.php
@@ -14,12 +14,17 @@ declare(strict_types=1);
 namespace Sylius\Bundle\ResourceBundle\DependencyInjection;
 
 use Sylius\Bundle\ResourceBundle\Controller\ResourceController;
+use Sylius\Bundle\ResourceBundle\DependencyInjection\Driver\Doctrine\DoctrineODMDriver;
+use Sylius\Bundle\ResourceBundle\DependencyInjection\Driver\Doctrine\DoctrineORMDriver;
+use Sylius\Bundle\ResourceBundle\DependencyInjection\Driver\Doctrine\DoctrinePHPCRDriver;
+use Sylius\Bundle\ResourceBundle\DependencyInjection\Driver\DriverInterface;
 use Sylius\Bundle\ResourceBundle\Form\Type\DefaultResourceType;
 use Sylius\Bundle\ResourceBundle\SyliusResourceBundle;
 use Sylius\Component\Resource\Factory\Factory;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
+use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 
 final class Configuration implements ConfigurationInterface
 {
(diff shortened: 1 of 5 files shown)


src/Bundle/DependencyInjection/Configuration.php
```diff
@@ -146,8 +151,76 @@ private function addDriversSection(ArrayNodeDefinition $node): void
...
+                                    unset($v[$driver]);
+
+                                    continue;
+                                }
+
+                                $v[$driver] = ['class' => $value];
+                            }
+
+                            return $v;
+                        })
```
loic425:
```suggestion
                        ->then(static function (array $values) {
                            foreach ($values as $driver => $value) {
                                if (isset($value['class'])) {
                                    continue;
                                }
                                
                                // bc-layer
                                if (in_array($value, SyliusResourceBundle::getAvailableDrivers(), true)) {
                                    $values[$value] = ['class' => match ($value) {
                                        SyliusResourceBundle::DRIVER_DOCTRINE_ORM => DoctrineORMDriver::class,
                                        SyliusResourceBundle::DRIVER_DOCTRINE_MONGODB_ODM => DoctrineODMDriver::class,
                                        SyliusResourceBundle::DRIVER_DOCTRINE_PHPCR_ODM => DoctrinePHPCRDriver::class,
                                    }];

                                    unset($values[$driver]);

                                    continue;
                                }

                                $values[$driver] = ['class' => $value];
                            }

                            return $values;
                        })
```

src/Bundle/DependencyInjection/Configuration.php
```diff
@@ -146,8 +151,76 @@ private function addDriversSection(ArrayNodeDefinition $node): void
...
+                    ->end()
+                    ->validate()
+                        ->ifArray()
+                        ->then(static function (array $v) {
+                            foreach ($v as $driver => $value) {
```
loic425, line 190: use "values" intead of "v"

src/Bundle/DependencyInjection/Driver/DriverProvider.php
```diff
@@ -13,41 +13,38 @@
...
      */
     public static function get(MetadataInterface $metadata): DriverInterface
     {
         $type = $metadata->getDriver();
 
-        if (isset(self::$drivers[$type])) {
-            return self::$drivers[$type];
+        /** @var class-string|null $class */
+        $class = self::$drivers[$type]['class'] ?? null;
+        if (null !== $class) {
```
loic425, line 40:
```suggestion

        if (null !== $class) {
```

src/Bundle/DependencyInjection/Driver/DriverProvider.php
```diff
@@ -13,41 +13,38 @@
...
+            $driver = new $class();
+            Assert::isInstanceOf($driver, DriverInterface::class);
```
loic425: Hum, that assert is on runtime, maybe it would be better to check that during configuration

src/Bundle/DependencyInjection/Driver/DriverProvider.php
```diff
@@ -13,41 +13,38 @@
...
     public static function get(MetadataInterface $metadata): DriverInterface
     {
         $type = $metadata->getDriver();
 
-        if (isset(self::$drivers[$type])) {
-            return self::$drivers[$type];
+        /** @var class-string|null $class */
+        $class = self::$drivers[$type]['class'] ?? null;
+        if (null !== $class) {
+            $driver = new $class();
```
loic425, line 42: Hum, That'd be better to use service instead of instantiating class, we can't have dependency injection in the drivers with your implementation.

src/Bundle/DependencyInjection/Driver/DriverProvider.php
Prometee: it is already, it was throwing an error during PSALM checks, I guess just adding a phpdoc will silent it

src/Bundle/DependencyInjection/Driver/DriverProvider.php
Prometee: We can't use a service because all the parameters created by the SyliusResourceBundle will have to be created during configuration. I don't know yet why this choice have been done, maybe there is a good reason.

loic425: @lchrusciel WDYT?

src/Bundle/DependencyInjection/Driver/DriverProvider.php
```diff
@@ -13,41 +13,39 @@
...
-use Sylius\Bundle\ResourceBundle\DependencyInjection\Driver\Doctrine\DoctrinePHPCRDriver;
 use Sylius\Bundle\ResourceBundle\DependencyInjection\Driver\Exception\UnknownDriverException;
-use Sylius\Bundle\ResourceBundle\SyliusResourceBundle;
 use Sylius\Component\Resource\Metadata\MetadataInterface;
 use Webmozart\Assert\Assert;
 
 final class DriverProvider
 {
-    /** @var DriverInterface[] */
+    /** @var array<string, array<string, string>> */
```
diimpp, line 22: I'm not too familiar with this construction, but will this work?

```patch
--- @var array<string, array<string, string>>
+++ @var array<string, array<string, class-string<DriverInterface>>>
```

Will it remove need to specify
```php
       /** @var class-string|null $class */
        $class = self::$drivers[$type]['class'] ?? null;
```
and
```php
           /** @var DriverInterface $driver */
            $driver = new $class();
```
later in the file?

src/Bundle/DependencyInjection/Driver/DriverProvider.php
```diff
@@ -13,41 +13,39 @@
...
      */
     public static function get(MetadataInterface $metadata): DriverInterface
     {
         $type = $metadata->getDriver();
 
-        if (isset(self::$drivers[$type])) {
-            return self::$drivers[$type];
-        }
+        /** @var class-string|null $class */
+        $class = self::$drivers[$type]['class'] ?? null;
```
diimpp, line 38: This line is somewhat misleading as `$class` is made nullable, while only `$drivers[$type]` is nullable and `$drivers[$type]['class']` should be asserted

```php
if (isset(self::$drivers[$type])) {
    Assert::notNull(self::$drivers[$type]['class'], sprintf('Class must be present for driver "%s" configuration', $type));
    $class = self::$drivers[$type]['class'];
    
    return new $class();
```

src/Bundle/DependencyInjection/Driver/DriverProvider.php
Prometee: Yes It works in this case, but the attribute php doc is here to define the generic type of a driver configuration. We should perhaps build a `DriverMetadata` class which own this configuration to better type it (I have to check if @loic425 already did this)

src/Bundle/DependencyInjection/Driver/DriverProvider.php
Prometee: IMHO this kind of assertion must be done into a `DriverMetadata` class like I suggest earlier.

PR #740, opened 2023-08-18
[Fix] Update delete path name to avoid route conflicts

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

1/ Delete path name can have some conflicts depending on operation sorting.
On previous routing system, these routing sorting system was hard-coded.
If we place the Bulk delete operation after the delete operation, the bulk delete is broken due to path conflicts.

2/ On current version of Symfony, http method overrides is not enabled by default and then these paths have another issue with same path for delete and edit operation (with POST method).
https://github.com/symfony/recipes/pull/892

Before
```
app_backend_contact_create               GET|POST        ANY      ANY    /admin/contacts/new                             
app_backend_contact_update               GET|PUT         ANY      ANY    /admin/contacts/{id}/edit                       
app_backend_contact_bulk_delete          DELETE          ANY      ANY    /admin/contacts/bulk_delete                     
app_backend_contact_delete               DELETE          ANY      ANY    /admin/contacts/{id}                          
app_backend_contact_show                 GET             ANY      ANY    /admin/contacts/{id}                            
app_backend_contact_index                GET             ANY      ANY    /admin/contacts 
```

After
```
app_backend_contact_create               GET|POST        ANY      ANY    /admin/contacts/new                             
app_backend_contact_update               GET|PUT         ANY      ANY    /admin/contacts/{id}/edit                       
app_backend_contact_bulk_delete          DELETE          ANY      ANY    /admin/contacts/bulk_delete                     
app_backend_contact_delete               DELETE          ANY      ANY    /admin/contacts/{id}/delete                           
app_backend_contact_show                 GET             ANY      ANY    /admin/contacts/{id}                            
app_backend_contact_index                GET             ANY      ANY    /admin/contacts 
```

### DIFF
--- a/src/Component/Symfony/Routing/Factory/DeleteOperationRoutePathFactory.php
+++ b/src/Component/Symfony/Routing/Factory/DeleteOperationRoutePathFactory.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sylius\Component\Resource\Symfony\Routing\Factory;
 
+use Sylius\Component\Resource\Metadata\Api\ApiOperationInterface;
 use Sylius\Component\Resource\Metadata\DeleteOperationInterface;
 use Sylius\Component\Resource\Metadata\HttpOperation;
 
@@ -28,10 +29,7 @@ final class DeleteOperationRoutePathFactory implements OperationRoutePathFactory
         $identifier = $operation->getResource()?->getIdentifier() ?? 'id';
 
         if ($operation instanceof DeleteOperationInterface) {
-            $path = match ($shortName) {
-                'delete' => '',
-                default => '/' . $shortName,
-            };
+            $path = $operation instanceof ApiOperationInterface && 'delete' === $shortName ? '' : '/' . $shortName;
 
             return sprintf('%s/{%s}%s', $rootPath, $identifier, $path);
         }
--- a/src/Component/Metadata/Api/ApiOperationInterface.php
+++ b/src/Component/Metadata/Api/ApiOperationInterface.php
@@ -1,0 +1,21 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Resource\Metadata\Api;
+
+/**
+ * @experimental
+ */
+interface ApiOperationInterface
+{
+}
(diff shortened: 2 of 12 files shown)


src/Component/Symfony/Routing/Factory/DeleteOperationRoutePathFactory.php
```diff
@@ -28,10 +29,7 @@ public function createRoutePath(HttpOperation $operation, string $rootPath): str
         $identifier = $operation->getResource()?->getIdentifier() ?? 'id';
 
         if ($operation instanceof DeleteOperationInterface) {
-            $path = match ($shortName) {
-                'delete' => '',
-                default => '/' . $shortName,
-            };
+            $path = $operation instanceof ApiOperationInterface && 'delete' === $shortName ? '' : '/' . $shortName;
```
Prometee, line 32: What about adding the `delete` string injected in this service, allowing developers to customize the url slug can maybe be a thing.

Prometee (approved):

src/Component/Symfony/Routing/Factory/DeleteOperationRoutePathFactory.php
loic425: Hum interesting, but if we do that for this one, we have to do that for every operations.

Prometee: It was just an idea, it will need a dedicated PR

lchrusciel, line 32: I have mixed feelings about it. The fix is needed, but in API this delete suffix should be required only in **non** API context. Once we are in API context, it is okay to have exactly the same URL but with different HTTP method. This logic is required only when we have a form, that will send a POST request with hidden DELETE notation. 

What I would recommend is to reverse this condition so we add `delete` only for non ApiOperationInterface

loic425: With my proposal, we have these behaviors.

**Non API**
```php
#[Delete]
#[Delete(name: 'remove')]
```

```
app_book_delete             DELETE        /admin/books/delete 
app_book_remove             DELETE        /admin/books/remove 
```

**API**
```php
#[Delete]
#[Delete(name: 'remove')]
```

```
app_book_delete             DELETE        /admin/books 
app_book_remove             DELETE        /admin/books/remove 
```

IMHO, this is ok like this.
Don't think about hidden field, cause we'll need to support POST method with non api routes.

With the next PR, it will be

**Non API**
```php
#[Delete]
#[Delete(name: 'remove')]
```

```
app_book_delete             POST, DELETE        /admin/books/delete 
app_book_remove             POST, DELETE        /admin/books/remove 
```

**API**
```php
#[Delete]
#[Delete(name: 'remove')]
```

```
app_book_delete             DELETE        /admin/books 
app_book_remove             DELETE        /admin/books/remove 
```

lchrusciel (approved):

lchrusciel:
Thank you, @loic425!

PR #757, opened 2023-10-20
Use AsResource attribute instead of reserved word Resource

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | could be one day
| New feature?    | no
| BC breaks?      | yes (but only on alpha version)
| Deprecations?   | no
| Related tickets | 
| License         | MIT

`resource` is a reserved word as Soyuka warned me on API CON.
So I think it will be better to use AsResource attribute instead.

if needed I can add an old Resource attribute with class_alias if we do not want to break users using alpha version.
Or maybe just add a note on the UPGRADE.md

I use a ResourceMetadata (old Resource one) object cause I don't want to make a `new AsResource()` call which is uggly.


lchrusciel (approved):

lchrusciel:
Thank you, @loic425!

PR #758, opened 2023-10-23
Remove Component namespace on action

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -100,7 +100,8 @@
     "autoload": {
         "psr-4": {
             "Sylius\\Bundle\\ResourceBundle\\": "src/Bundle/",
-            "Sylius\\Component\\Resource\\": "src/Component/"
+            "Sylius\\Component\\Resource\\": "src/Component/",
+            "Sylius\\Resource\\": "src/Component/src/"
         }
     },
     "autoload-dev": {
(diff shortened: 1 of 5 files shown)


src/Component/Action/PlaceHolderAction.php
```diff
@@ -13,13 +13,10 @@
...
-final class PlaceHolderAction
-{
-    /**
-     * @param mixed $data
-     */
-    public function __invoke($data = null): mixed
+class_exists(\Sylius\Resource\Action\PlaceHolderAction::class);
+
+if (false) {
+    final class PlaceHolderAction extends \Sylius\Resource\Action\PlaceHolderAction
```
diimpp: Is it alright to extend `final` class?

diimpp (approved):

src/Component/Action/PlaceHolderAction.php
loic425: Yes, it's strange but it's done like this. I already did that based on API Platform bc layer and it works.

lchrusciel (commented):
I'm all in favour of this change about the namespace, but there is no point in doing a migration layer for something that didn't reach official release

loic425:
> I'm all in favour of this change about the namespace, but there is no point in doing a migration layer for something that didn't reach official release

Indeed, but I prefer to use the same bc-layer for every moved classes/interfaces to avoid errors (even if they are new).

lchrusciel (approved):

lchrusciel:
Thank you, @loic425!

PR #777, opened 2023-11-08
Move some namespaces on Factory

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT


src/Component/Factory/TranslatableFactoryInterface.php
diimpp, line 1: I think this file needs BC layer as for other moved classes/interfaces.

src/Component/spec/Factory/FactorySpec.php
diimpp, line 1: Shouldn't this file be moved into `src/Component/tests/spec`?

src/Component/spec/Factory/TranslatableFactorySpec.php
diimpp, line 1: Shouldn't this file be moved into `src/Component/tests/spec`?

src/Component/src/Factory/Factory.php
```diff
@@ -0,0 +1,41 @@
...
+final class Factory implements FactoryInterface
+{
+    /**
+     * @var string
+     * @psalm-var class-string
+     */
+    private $className;
+
+    /**
+     * @psalm-param class-string $className
```
diimpp: Outside of scope of current PR,

Those `psalm-param` and `psalm-var` looks like a legacy solution for times when phpstan didn't have class-string and it can be updated now. (I don't have strict knowledge about this, just a feeling)

src/Component/Factory/TranslatableFactoryInterface.php
loic425: Hum I added the class alias and class exists So I think it's ok. But cause this factory is very important, please tell me if I forgot something you think about.

src/Component/spec/Factory/FactorySpec.php
loic425: Yeah, I'm just wondering if I just duplicate the spec to be sure the alias is ok.

loic425: or maybe just testing the alias is ok....

src/Component/Factory/TranslatableFactoryInterface.php
loic425: Oh god you're right, I was missing the interface alias

diimpp (approved):

src/Component/spec/Factory/FactorySpec.php
diimpp: Cool idea :)

oallain (approved):

lchrusciel (approved):

lchrusciel:
Thank you, @loic425!

PR #780, opened 2023-11-10
Init Deserialize provider

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Needs #781 first

Not used and no DI for now, cause we need the ReadProvider first.

PR #782, opened 2023-11-10
Phpunit tests for Symfony read listener

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Based on #780


src/Component/Tests/Symfony/EventListener/ReadListenerTest.php
```diff
@@ -0,0 +1,167 @@
...
+
+        $operation = new Create();
+
+        $context = new Context();
+
+        $this->contextInitiator->initializeContext($request)->willReturn($context);
+
+        $this->operationInitiator->initializeOperation($request)->willReturn($operation);
+
+        $request->attributes = $attributes;
```
diimpp, line 99: Is it possible or makes sense to prophesize attributes directly to `$request->attributes`?

diimpp (approved):

src/Component/Tests/Symfony/EventListener/ReadListenerTest.php
loic425: nope, we cannot do that.

loic425: We also did like that on the old spec file.

PR #783, opened 2023-11-13
Move some namespaces on Symfony forms, requests and responses

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

These classes have been introduced on 1.11.


src/Bundle/Resources/config/services/state.xml
```diff
@@ -15,22 +15,6 @@
     <services>
         <service id="sylius.resource_factory.expression_language" class="Symfony\Component\ExpressionLanguage\ExpressionLanguage" />
 
-        <service id="sylius.resource_factory.argument_parser" class="Sylius\Component\Resource\Factory\ArgumentParser">
```
loic425, line 18: this one does not existing anymore

src/Bundle/Resources/config/services/state.xml
```diff
@@ -15,22 +15,6 @@
...
-            <argument type="service" id="sylius.resource_factory.expression_language" />
-            <argument type="service" id="request_stack" />
-            <argument type="service" id="security.token_storage" on-invalid="null" />
-        </service>
-        <service id="Sylius\Component\Resource\Factory\ArgumentParserInterface" alias="sylius.resource_factory.argument_parser" />
-
-        <service id="sylius.state.provider.event_dispatcher"
```
loic425, line 25: this service configuration was duplicated

src/Bundle/Resources/config/services/form.xml
```diff
@@ -12,9 +12,9 @@
         </service>
         <service id="Sylius\Bundle\ResourceBundle\Form\Type\ResourceAutocompleteChoiceType" alias="sylius.form.type.resource_autocomplete_choice" />
 
-        <service id="sylius.form.factory" class="Sylius\Component\Resource\Symfony\Form\Factory\FormFactory" public="false">
+        <service id="sylius.form.factory" class="Sylius\Resource\Symfony\Form\Factory\FormFactory" public="false">
             <argument type="service" id="form.factory" />
         </service>
-        <service id="Symfony\Component\Form\FormFactoryInterface" alias="sylius.form.factory" />
+        <service id="Sylius\Resource\Symfony\Form\Factory\FormFactoryInterface" alias="sylius.form.factory" />
```
diimpp, line 18: It was introduced in 1.11, right? So it doesn't need a BC.

Strange that SF `FormFactoryInterface` and not sylius interface was aliased though.

diimpp (approved):

src/Bundle/Resources/config/services/form.xml
loic425: Yes, it was introduced on 1.11.
And yes this is very strange, this was my error. Sometimes, using completion could be dangerous :)

lchrusciel:
Thank you, @loic425!

PR #785, opened 2023-11-16
Init read provider

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

Based on #782


src/Component/src/State/Provider/ReadProvider.php
```diff
@@ -0,0 +1,51 @@
...
+            return null;
+        }
+
+        $data = $this->provider->provide($operation, $context);
+
+        if (null === $data) {
+            throw new NotFoundHttpException('Resource has not been found.');
+        }
+
+        $request?->attributes->set('data', $data);
```
diimpp, line 47: Is `$request?->` due static analyze strictness or there are cases, where request doesn't have attributes?

diimpp (approved):

src/Component/src/State/Provider/ReadProvider.php
loic425: we can use this read provider without any HTTP request :)
So there are cases when request is null.

lchrusciel (approved):

lchrusciel:
Thank you, @loic425!

PR #805, opened 2023-11-29
Move some namespaces on generator

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Based on #804

PR #806, opened 2023-11-30
Move some namespaces on reflection

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Based on #805

Only ClassReflection needs a bc-layer cause the other classes have been introduced in 1.11.


src/Component/Reflection/ClassReflection.php
```diff
@@ -13,58 +13,11 @@
...
-     */
-    public static function getClassAttributes(string $className, ?string $attributeName = null): array
-    {
-        $reflectionClass = new \ReflectionClass($className);
-
-        /** @psalm-suppress ArgumentTypeCoercion */
-        return $reflectionClass->getAttributes($attributeName);
     }
 }
+
```
diimpp: Judging by diff, there is unnecessary empty line at the end.

But in file view it's alright :thinking: 
image

src/Component/src/Reflection/ClassReflection.php
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    public static function getResourcesByPath(string $path): iterable
+    {
+        $finder = new Finder();
+        $finder->files()->in($path)->name('*.php')->sortByName(true);
+
+        foreach ($finder as $file) {
+            $fileContent = (string) file_get_contents((string) $file->getRealPath());
```
diimpp:
```suggestion
            $fileContent = file_get_contents((string) $file->getRealPath());
```

`file_get_contents` might return **false**, so it might be better to handle it explicitly.

diimpp (approved):

src/Component/src/Reflection/ClassReflection.php
loic425: Done

src/Component/Reflection/ClassReflection.php
loic425: done

lchrusciel:
Thank you, @loic425!

PR #809, opened 2023-12-01
Add support for Symfony 6.4

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes (cause we do not exclude it on composer)
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -71,6 +71,7 @@
         "sylius/grid-bundle": "^1.7 || v1.12.0-ALPHA.1",
         "symfony/dependency-injection": "^5.4 || ^6.0",
         "symfony/dotenv": "^5.4 || ^6.0",
+        "symfony/stopwatch": "^5.4 || ^6.0",
         "symfony/workflow": "^5.4 || ^6.0",
         "twig/twig": "^2.12 || ^3.0",
         "vimeo/psalm": "^4.22",
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,16 +18,16 @@ jobs:
             fail-fast: false
             matrix:
                 php: ["8.0", "8.1"]
-                symfony: ["^5.4", "~6.0.0", "~6.1.0", "~6.2.0"]
+                symfony: ["^5.4", "~6.0.0", "~6.4.0"]
                 twig: ["^2.12", "^3.0"]
                 persistence: ["^2.0", "^3.0"]
                 exclude:
                     -
                         php: "8.0"
-                        symfony: "~6.1.0"
+                        symfony: "~6.4.0"
                     -
-                        php: "8.0"
-                        symfony: "~6.2.0"
+                        persistence: "^2.0"
+                        symfony: "~6.4.0"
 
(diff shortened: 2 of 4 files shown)


src/Bundle/Controller/ContainerAwareTrait.php
```diff
@@ -0,0 +1,34 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ResourceBundle\Controller;
+
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * @internal
+ */
+trait ContainerAwareTrait
```
loic425: I've copied the Symfony trait cause it's now deprecated

composer.json
```diff
@@ -71,6 +71,7 @@
         "sylius/grid-bundle": "^1.7 || v1.12.0-ALPHA.1",
         "symfony/dependency-injection": "^5.4 || ^6.0",
         "symfony/dotenv": "^5.4 || ^6.0",
+        "symfony/stopwatch": "^5.4 || ^6.0",
```
loic425, line 74: I've added this cause there is a bug with workflow without stopwatch.
These 2 ones are on dev requirements

src/Bundle/Controller/ContainerAwareTrait.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
```
GSadee:
```suggestion
 * (c) Sylius Sp. z o.o.
```

src/Bundle/Controller/ContainerAwareTrait.php
GSadee: It is deprecated but not removed in 6.4, so do we really need to copy it in this version? 🤔

tests/Application/config/db.sql
GSadee, line 1: Is this file added intentionally? Because I don't see the connection with this PR 🤔

loic425: oups...

src/Bundle/Controller/ContainerAwareTrait.php
loic425: Yeah but it's removed on 7.0... But right, maybe adding it in 1.11 (the current development one which contains the new resource metadata system).

src/Bundle/Controller/ContainerAwareTrait.php
loic425: We did this on the 1.11 branch, but not on the 1.10 (for bugfix)

diimpp (approved):

diimpp:
@loic425 1.10 is not supported anymore, it should target 1.11

image

GSadee (approved):

GSadee:
Thank you, Loïc! :tada:

PR #819, opened 2023-12-11
Move some namespaces on model

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT


src/Bundle/EventListener/ODMTranslatableListener.php
```diff
@@ -68,11 +70,11 @@ public function loadClassMetadata(LoadClassMetadataEventArgs $eventArgs)
             return;
         }
 
-        if ($reflection->implementsInterface(TranslatableInterface::class)) {
+        if ($reflection->implementsInterface(TranslatableInterface::class) || $reflection->implementsInterface(LegacyTranslatableInterface::class)) {
```
loic425: Hum, I'm not sure we need that.

tests/Application/src/Entity/LegacyBook.php
```diff
@@ -0,0 +1,44 @@
...
+    use TranslatableTrait {
+        __construct as private initializeTranslationsCollection;
+    }
+
+    #[ORM\Id]
+    #[ORM\GeneratedValue]
+    #[ORM\Column]
+    private int $id;
+
+    public function getId(): int
```
diimpp:
```suggestion
    private ?int $id = null;

    public function getId(): ?int
```

Current typehint will throw exception on call on new entity.

diimpp: Would be cool to sync code with regular Book/BookTranslation

> tests/Application/src/Entity/Book.php

diimpp (approved):

lchrusciel:
Thank you, @loic425!

PR #817, opened 2023-12-08
Add DI for Main controller, providers and processors

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

On browsing resources
Capture d’écran de 2023-12-15 16-02-41

On creating a new resource
Capture d’écran de 2023-12-15 16-03-18

On invalid creation
Capture d’écran de 2023-12-15 16-04-16

On deleting a single resource
Capture d’écran de 2023-12-15 16-05-01

On delete multiple resources
Capture d’écran de 2023-12-15 16-00-59

PR #824, opened 2023-12-16
Add main controller

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Component/src/Symfony/Controller/MainController.php
+++ b/src/Component/src/Symfony/Controller/MainController.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Sylius\Resource\Symfony\Controller;
 
 use Sylius\Resource\Context\Initiator\RequestContextInitiatorInterface;
+use Sylius\Resource\Exception\RuntimeException;
 use Sylius\Resource\Metadata\Operation\HttpOperationInitiatorInterface;
 use Sylius\Resource\State\ProcessorInterface;
 use Sylius\Resource\State\ProviderInterface;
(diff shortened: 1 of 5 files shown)


src/Component/src/Symfony/Controller/MainController.php
```diff
@@ -0,0 +1,56 @@
...
+        if (null === $operation->canWrite()) {
+            $operation = $operation->withWrite(!$request->isMethodSafe());
+        }
+
+        $data = $this->provider->provide($operation, $context);
+
+        $valid = $request->attributes->getBoolean('is_valid', true);
+        if (!$valid) {
+            $operation = $operation->withWrite(false);
+        }
```
lchrusciel, line 53:
```suggestion
        $operation = $operation->withWrite($request->attributes->getBoolean('is_valid', true));
```

But I wonder if `is_valid` should not be false by default

lchrusciel (approved):

src/Component/src/Symfony/Controller/MainController.php
loic425: If we configure "validate: false" on an operation, we don't have this attribute on request.

diimpp (commented):
Left few comments on previous PR https://github.com/Sylius/SyliusResourceBundle/pull/817#pullrequestreview-1785322834

src/Component/src/Symfony/Controller/MainController.php
loic425: I do not agree with your suggestion, the developer could configure Create(write: false), validation is ok, and then it will write and developer does not want that.

diimpp (approved):

lchrusciel:
Thank you, @loic425!

PR #828, opened 2023-12-18
Dispatch pre write event processor

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

It was PreEventDispatcherProcessor but I think it's a better name now.

I've decided to split https://github.com/Sylius/SyliusResourceBundle/blob/1.11/src/Component/src/State/Processor/EventDispatcherProcessor.php into two processors, so this one is the first part of the split.

This processor allows user switching from resource controller to this new system easily.
But we will be able to deprecate this event in the future.

### DIFF
--- a/src/Component/src/Symfony/EventDispatcher/State/DispatchPreWriteEventProcessor.php
+++ b/src/Component/src/Symfony/EventDispatcher/State/DispatchPreWriteEventProcessor.php
@@ -1,0 +1,55 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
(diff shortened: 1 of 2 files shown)


diimpp (approved):

oallain (approved):

lchrusciel (approved):
However I wonder if this single processor shouldn't be responsible for both pre and post events

lchrusciel:
Thank you, @loic425!

PR #829, opened 2023-12-18
Dispatch post write event processor

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Based on https://github.com/Sylius/SyliusResourceBundle/pull/828

### DIFF
--- a/src/Component/src/Symfony/EventDispatcher/State/DispatchPostWriteEventProcessor.php
+++ b/src/Component/src/Symfony/EventDispatcher/State/DispatchPostWriteEventProcessor.php
@@ -1,0 +1,54 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Resource\src\Symfony\EventDispatcher\State;
+
+use Sylius\Resource\Context\Context;
+use Sylius\Resource\Metadata\Operation;
+use Sylius\Resource\State\ProcessorInterface;
+use Sylius\Resource\Symfony\EventDispatcher\OperationEventDispatcherInterface;
+use Sylius\Resource\Symfony\EventDispatcher\OperationEventHandlerInterface;
+
+/**
+ * @experimental
+ */
+final class DispatchPostWriteEventProcessor implements ProcessorInterface
+{
+    public function __construct(
+        private ProcessorInterface $processor,
+        private OperationEventDispatcherInterface $operationEventDispatcher,
+        private OperationEventHandlerInterface $eventHandler,
+    ) {
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


src/Component/src/Symfony/EventDispatcher/State/DispatchPostWriteEventProcessor.php
```diff
@@ -0,0 +1,54 @@
...
+        private OperationEventHandlerInterface $eventHandler,
+    ) {
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function process(mixed $data, Operation $operation, Context $context): mixed
+    {
+        $data = $this->processor->process($data, $operation, $context);
```
diimpp, line 39: Is it alright for both `PreWrite` and `PostWrite` to call "inner" processor? Won't it result on double execution?

src/Component/src/Symfony/EventDispatcher/State/DispatchPostWriteEventProcessor.php
```diff
@@ -0,0 +1,54 @@
...
+
+        $operationEvent = $this->operationEventDispatcher->dispatchPostEvent($data, $operation, $context);
```
diimpp, line 41: From naming point of view, does `$this->operationEventDispatcher->dispatchPostEvent` dispatches it for listeners to listen or it's only a event object factory?

Generally speaking, I cannot understand what's going on with "dispatch event" and "handle event" in two consequent lines of code without reading internals of both.

diimpp (approved):

src/Component/src/Symfony/EventDispatcher/State/DispatchPostWriteEventProcessor.php
loic425: No, I've checked and that was ok. The harder thing is to create tests for that and I have no idea.

src/Component/src/Symfony/EventDispatcher/State/DispatchPostWriteEventProcessor.php
loic425: Yeah I do not like this part, and that's only to allow developers to switch from resource controller to new operations system easily.

On resource controller:
Pre write
https://github.com/Sylius/SyliusResourceBundle/blob/1.11/src/Bundle/Controller/ResourceController.php#L175
Post write
https://github.com/Sylius/SyliusResourceBundle/blob/1.11/src/Bundle/Controller/ResourceController.php#L202

loic425: In the future, the idea is to handle this with your custom processor or decorating a Sylius one in the e-commerce part when Sylius will use operations.

oallain (approved):

GSadee (approved):

GSadee:
Thanks, Loïc! :tada:

PR #836, opened 2024-01-11
Remove cache on metadata when debug is enabled

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

On dev, when updating metadata, the metadata (and so the routing) was in cache due to this service. But on dev/test environment, we don't need that cache, we need to update the routing while configuring the attributes on our Sylius resources.


diimpp (approved):

oallain (approved):

GSadee (approved):

GSadee:

PR #841, opened 2024-01-19
Add support for PHP 8.3 (CI)

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

There is still one test to include php8.0.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,8 +27,6 @@ jobs:
                       pagerfanta: "^3.7"
                       symfony: "^5.4"
                 exclude:
-                    - php: "8.0"
-                      collections: "^2.0"
                     - php: "8.1"
                       collections: "^2.0"
                       persistence: "^2.0"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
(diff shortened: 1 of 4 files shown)


diimpp (approved):

oallain (approved):

.github/workflows/build.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: ["8.0", "8.1", "8.2"]
+                php: ["8.1", "8.2", "8.3"]
```
GSadee, line 20: As we do not bump up the PHP version in composer, I would be in favour of still testing 8.0. Preferably the lowest (8.0) and highest (8.3) versions if we want to reduce the number of builds in matrix and add single builds for 8.1 and 8.2 in `include`.

loic425: that proposal uses the opposite, we still have a single build with php 8.0
https://github.com/Sylius/SyliusResourceBundle/blob/0f937de583b1d304dfb19d200da1f34bcaf45e9a/.github/workflows/build.yml#L25

GSadee (approved):

GSadee:
Thanks, Loïc! :tada:

PR #852, opened 2024-02-16
Refactor component directories

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT


src/Component/spec/Model/AbstractTranslationSpec.php
```diff
@@ -11,28 +11,48 @@
 
 declare(strict_types=1);
 
-namespace spec\Sylius\Component\Resource\Model;
+namespace spec\Sylius\Resource\Model;
```
loic425, line 14: no change here, the legacy file has been moved and the new one has been placed here.

src/Component/spec/Metadata/RegistrySpec.php
```diff
@@ -11,11 +11,10 @@
 
 declare(strict_types=1);
 
-namespace spec\Sylius\Component\Resource\Metadata;
+namespace spec\Sylius\Resource\Metadata;
```
loic425, line 14: Same here.

phpspec.yml.dist
```diff
@@ -7,13 +7,13 @@ suites:
...
-        src_path: src/Component
-        spec_path: src/Component
+        src_path: src/Component/legacy
+        spec_path: src/Component/legacy
     Package:
         namespace: Sylius\Resource
         psr4_prefix: Sylius\Resource
         src_path: src/Component/src
-        spec_path: src/Component/tests
+        spec_path: src/Component
```
diimpp, line 16: Just in case, `Package` suite has `src` in path and `Component` doesn't.

```yml
        src_path: src/Component/legacy
        spec_path: src/Component/legacy

        src_path: src/Component/src
        spec_path: src/Component
```

src/Component/phpunit.xml.dist
```diff
@@ -6,7 +6,7 @@
 >
     <testsuites>
         <testsuite name="SyliusResource Test Suite">
-            <directory>./Tests/</directory>
+            <directory>./legacy/tests/</directory>
```
diimpp, line 9: Just in case, non-legacy tests are loaded as well?

image

src/Component/spec/Generator/RandomnessGeneratorSpec.php
```diff
@@ -25,13 +23,41 @@ function it_implements_randomness_generator_interface(): void
...
+        $this->generateInt($min, $max)->shouldBeInt();
+        $this->generateInt($min, $max)->shouldBeInRange($min, $max);
     }
 
-    function it_should_be_an_alias_of_randomness_generator(): void
+    public function getMatchers(): array
     {
-        $this->shouldBeAnInstanceOf(NewRandomnessGenerator::class);
+        return [
+            'haveLength' => function ($subject, $length) {
```
diimpp: Is it applicable over here?
```suggestion
            'haveLength' => function (string $subject, int $length): bool {
```

src/Component/spec/Generator/RandomnessGeneratorSpec.php
```diff
@@ -25,13 +23,41 @@ function it_implements_randomness_generator_interface(): void
...
+                return $length === strlen($subject);
+            },
+            'beInRange' => function ($subject, $min, $max) {
```
diimpp:
```suggestion
            'beInRange' => function (int $subject, int $min, int $max): bool {
```

diimpp (approved):
At this point, why not move `Sylius\Resource` to `src/Resource` instead of keeping it at `src/Component`?

src/Component/spec/Metadata/MetadataSpec.php
```diff
@@ -11,27 +11,157 @@
...
+        $this->getServiceId('factory')->shouldReturn('app.factory.product');
+        $this->getServiceId('repository')->shouldReturn('app.repository.product');
+        $this->getServiceId('form.type')->shouldReturn('app.form.type.product');
     }
 
-    function it_should_be_an_alias_of_metadata(): void
+    function it_generates_permission_code(): void
     {
-        $this->shouldBeAnInstanceOf(NewMetadata::class);
+        $this->getPermissionCode('show')->shouldReturn('app.product.show');
```
diimpp, line 163: Shouldn't it be based on resource name?
```suggestion
        $this->getPermissionCode('show')->shouldReturn('app.product.with.dots.show');
```

src/Component/phpunit.xml.dist
loic425: yes, they are. To be honest, this phpunit file is not used (don't know since when). But I think I will use it when we'll get rid of the bundle.

phpspec.yml.dist
loic425: hum... strange, I need to check

loic425: LOL, that works in both cases, but I've changed that.

src/Component/spec/Metadata/MetadataSpec.php
loic425: in the let method, the product has "'app.product'" alias.

lchrusciel (approved):
But I also like Dmitri idea 👍

lchrusciel:
Thanks @loic425 🏅

PR #871, opened 2024-05-14
Fix state machine tag

Zales0123:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

👋 As Sylius introduced the state machine layer abstraction with its own `sylius.state_machine` service tag, we have a little conflict of configuration, which prevents using SyliusResourceBundle with Sylius 1.13 for now :dancer: (see [this issue](https://github.com/Sylius/Sylius/issues/16231))

For now, I propose we change the tag in this repository to `sylius_resource.state_machine` (which is also semantically more correct), but we should probably think if there is any thing we need to do on this repo according to this state machine abstraction 🖖

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -120,6 +120,13 @@
             </errorLevel>
         </MethodSignatureMismatch>
 
+        <MissingClassConstType>
+            <errorLevel type="suppress">
+                <directory name="src/Bundle" />
+                <directory name="src/Component" />
+            </errorLevel>
+        </MissingClassConstType>
+
         <MissingParamType>
             <errorLevel type="suppress">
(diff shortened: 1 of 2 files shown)


lchrusciel (approved):

lchrusciel:
Thank you, @Zales0123!

GSadee:
Adding to the context, in Sylius/Sylius, the state machine from Resource is overridden to use abstraction (https://github.com/Sylius/Sylius/pull/16051), so there is probably no need for any other changes here. Unless we would like to use the StateMachineAbstraction package here 🤔

PR #872, opened 2024-05-15
Move fos rest, jms serializer and hateoas on optional requirements

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | mentioned in upgrade
| Deprecations?   | no
| Related tickets | 
| License         | MIT

It's time to get rid of these packages. Let's start by moving them on optional requirements.

- [x] move on dev dependencies
- [x] fix DI
- [x] FosRestPassTest
- [x] HateoasPassTest

PR #873, opened 2024-05-15
Move winzou state machine on optional requirements

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | mentioned on UPGRADE.md
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Based on #872


Wojdylak (approved):

composer.json
```diff
@@ -86,14 +85,16 @@
...
+        "willdurand/hateoas-bundle": "^2.0",
+        "winzou/state-machine-bundle": "^0.6.2"
     },
     "conflict": {
         "doctrine/orm": "<2.18 || ^3.0",
         "doctrine/doctrine-bundle": "<2.0 || ^3.0",
         "friendsofsymfony/rest-bundle": "<3.0",
         "jms/serializer-bundle": "<3.5",
         "willdurand/hateoas-bundle": "<2.0 || ^2.6",
+        "winzou/state-machine-bundle": "<0.6.2",
```
Rafikooo, line 97: Is this necessary if we have specified the minimum version as `"winzou/state-machine-bundle": "^0.6.2"`?

Rafikooo (approved):

composer.json
GSadee: We have, but only in dev-requirement, so to be sure in the end application that the proper and supported version will be installed, this conflict is needed

GSadee (approved):

PR #875, opened 2024-05-17
Implementing repository interface is not mandatory anymore

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | not really
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

### DIFF
--- a/docs/create_new_resource.md
+++ b/docs/create_new_resource.md
@@ -124,10 +124,9 @@ class BookRepository extends ServiceEntityRepository
 }
 ```
 
-The generated code is not compatible with Sylius Resource yet, so we need to make few changes.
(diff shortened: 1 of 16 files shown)


docs/create_new_resource.md
```diff
@@ -124,10 +124,9 @@ class BookRepository extends ServiceEntityRepository
 }
 ```
 
-The generated code is not compatible with Sylius Resource yet, so we need to make few changes.
+The generated code could be not compatible with Sylius Resource in some cases, so we need to make few changes.
```
loic425, line 127: Hum I think I should move all this part on Index operation without any grid.

PR #881, opened 2024-05-23
[SF7] Remove enable authenticator manager option on test application

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes (for sf7 test application)
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Based on #885

Before
image


composer.json
```diff
@@ -70,20 +70,21 @@
         "phpstan/phpstan-phpunit": "^1.1",
         "phpstan/phpstan-webmozart-assert": "^1.2",
         "phpunit/phpunit": "^9.5",
+        "rector/rector": "^0.18.2",
         "sylius-labs/coding-standard": "^4.0",
         "sylius/grid-bundle": "^1.7 || dev-symfony-7",
         "symfony/console": "^5.4 || ^6.4 || ^7.0",
         "symfony/dependency-injection": "^5.4 || ^6.4 || ^7.0",
         "symfony/dotenv": "^5.4 || ^6.4 || ^7.0",
+        "symfony/http-kernel": "^5.4 || ^6.4 || ^7.0",
```
loic425, line 79: this is the only one dependency I've added on dev requirements, the others are just moved up or down.

composer.json
```diff
@@ -70,20 +70,21 @@
         "phpstan/phpstan-phpunit": "^1.1",
         "phpstan/phpstan-webmozart-assert": "^1.2",
         "phpunit/phpunit": "^9.5",
+        "rector/rector": "^0.18.2",
```
diimpp, line 73: Should it depend on https://github.com/Sylius/SyliusRector instead? Then explicit `rector/rector` requirement is not needed anymore.

I'm not fully sure why syliusrector exists though.

diimpp (approved):

composer.json
loic425: normally this kind of things is already ok if we use Docker or CI to set up the dev requirements.
If you don't use docker it will not use the "symfony requirement from Flex" filter if we don't have flex on composer globals.

loic425: but I think it's better to explicit all our dev dependencies.

composer.json
loic425: hum sylius rector was exists to add some custom rules to bump Sylius e-commerce versions. But I think it's not used anymore.

PR #888, opened 2024-05-23
[SF7] Fix the branch build

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes, on that branch
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,8 @@ COPY . /app
 
 WORKDIR /app
 
+RUN composer global config --no-plugins allow-plugins.symfony/flex true
+RUN composer global require --no-progress --no-scripts --no-plugins "symfony/flex:^1.10"
 RUN composer update --with-all-dependencies --no-interaction --no-progress
 
 WORKDIR /app/tests/Application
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -104,6 +104,9 @@
     "extra": {
         "branch-alias": {
             "dev-master": "1.10-dev"
+        },
+        "symfony": {
+            "require": "^6.4"
```
loic425, line 109: it allow to fix the docker dev build

Dockerfile
```diff
@@ -15,6 +15,8 @@ COPY . /app
 
 WORKDIR /app
 
+RUN composer global config --no-plugins allow-plugins.symfony/flex true
```
loic425, line 18: it allow to fix the docker dev build

composer.json
loic425: With that configuration, it will set up dependencies with Symfony 6.4.

Dockerfile
loic425: we do not require flex in dev requirements, cause when a developer run "composer install" it will apply flex recipes.

diimpp (approved):

PR #890, opened 2024-05-23
[SF7] Fix resource loader

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes on sf7
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Based on #887 

image

### DIFF
--- a/src/Bundle/Routing/ResourceLoader.php
+++ b/src/Bundle/Routing/ResourceLoader.php
@@ -17,22 +17,24 @@ use Gedmo\Sluggable\Util\Urlizer;
 use Sylius\Resource\Metadata\MetadataInterface;
 use Sylius\Resource\Metadata\RegistryInterface;
 use Symfony\Component\Config\Definition\Processor;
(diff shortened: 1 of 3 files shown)


src/Bundle/Routing/ResourceLoader.php
```diff
@@ -105,17 +107,15 @@ public function supports($resource, $type = null): bool
 
     /**
      * @psalm-suppress InvalidReturnType Symfony docblocks are messing with us
-     *
-     * @return LoaderResolverInterface
      */
-    public function getResolver()
+    public function getResolver(): LoaderResolverInterface
```
diimpp: Since it will crash if `getResolver` will be called before `setResolver`, let's handle it the same as in symfony.

https://github.com/symfony/symfony/blob/bed9c3660c0106049622eead422fcb3921f8dd21/src/Symfony/Component/Config/Loader/Loader.php#L24-L37

diimpp (approved):

PR #891, opened 2024-05-23
[SF7] Fix analysis errors

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes on SF7 branch
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Bundle/Form/DataTransformer/ResourceToIdentifierTransformer.php
+++ b/src/Bundle/Form/DataTransformer/ResourceToIdentifierTransformer.php
@@ -35,7 +35,7 @@ final class ResourceToIdentifierTransformer implements DataTransformerInterface
     /**
      * @inheritDoc
      */
-    public function transform($value): mixed
+    public function transform(mixed $value): mixed
     {
         if (null === $value) {
             return null;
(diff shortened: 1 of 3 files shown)


src/Bundle/Form/DataTransformer/ResourceToIdentifierTransformer.php
```diff
@@ -33,13 +33,9 @@ public function __construct(RepositoryInterface $repository, ?string $identifier
...
     /**
-     * @psalm-suppress MissingParamType
-     *
-     * @param object|null $value
-     *
-     * @return mixed
+     * @inheritDoc
      */
-    public function transform($value)
+    public function transform($value): mixed
```
loic425: This is a final class

diimpp:
```suggestion
    public function transform(mixed $value): mixed
```
https://github.com/symfony/symfony/blob/bed9c3660c0106049622eead422fcb3921f8dd21/src/Symfony/Component/Form/DataTransformerInterface.php#L65

diimpp:
```php
    /** @param int|string|null $value */
public function reverseTransform($value): ?ResourceInterface
```
Not sure how it's even working, since correct typehint is 
```php
 public function reverseTransform(mixed $value): mixed;
```

loic425: you mean for the return typehint? this is ok, you can restrict it but I don't even remember on which php version.

diimpp (approved):

PR #924, opened 2024-09-05
Remove Twig 2 support

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -83,14 +83,15 @@
         "symfony/messenger": "^5.4 || ^6.4 || ^7.0",
         "symfony/serializer": "^5.4 || ^6.4 || ^7.0",
         "symfony/security-bundle": "^5.4 || ^6.4 || ^7.0",
-        "twig/twig": "^2.12 || ^3.0",
+        "twig/twig": "^3.0",
         "vimeo/psalm": "^5.20",
         "willdurand/hateoas-bundle": "^2.0"
     },
     "conflict": {
         "friendsofsymfony/rest-bundle": "<3.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -83,14 +83,15 @@
...
+        "twig/twig": "^3.0",
         "vimeo/psalm": "^5.20",
         "willdurand/hateoas-bundle": "^2.0"
     },
     "conflict": {
         "friendsofsymfony/rest-bundle": "<3.0",
         "jms/serializer-bundle": "<3.5",
-        "willdurand/hateoas-bundle": "<2.0 || ^2.6"
+        "willdurand/hateoas-bundle": "<2.0 || ^2.6",
+        "twig/twig": "<3.0"
```
GSadee, line 94: As we have bumped up the dependency, I'm not sure if it is necessary to add this package additionally to the conflicts (and the about note to the `CONFLICTS` file), do you see any benefit of that? 🤔

loic425: cause it's on require-dev, it needs a conflict. We often forgot about those conflicts when there are on "optional/dev" requirements.
However, I do not know if adding note in the conflict make sense, cause it's just a bumped dependency.

loic425: I've removed CONFLICTS file part.

GSadee (approved):

GSadee:
Thank you, Loïc! :tada:

PR #926, opened 2024-09-09
[Composer] Allow GridBundle ^v1.13@alpha

GSadee:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no 
| Related tickets | 
| License         | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -72,7 +72,7 @@
         "phpunit/phpunit": "^9.5",
         "rector/rector": "^0.18.2",
         "sylius-labs/coding-standard": "^4.0",
-        "sylius/grid-bundle": "^1.7 || dev-symfony-7",
+        "sylius/grid-bundle": "^1.7 || ^1.13@alpha",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -72,7 +72,7 @@
         "phpunit/phpunit": "^9.5",
         "rector/rector": "^0.18.2",
         "sylius-labs/coding-standard": "^4.0",
-        "sylius/grid-bundle": "^1.7 || dev-symfony-7",
+        "sylius/grid-bundle": "^1.7 || ^v1.13@alpha",
```
loic425: I don't think "v" prefix is needed.

loic425: I think we should bump that grid bundle version cause this one is not maintained anymore. (don't forget to add the conflict if you do).

loic425 (approved):

PR #934, opened 2024-09-11
Reproduce getting the translation repository by its interface use case

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

On 1.12 branch 
image

### DIFF
--- a/tests/Application/src/Entity/BookTranslationInterface.php
+++ b/tests/Application/src/Entity/BookTranslationInterface.php
@@ -1,0 +1,23 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace App\Entity;
+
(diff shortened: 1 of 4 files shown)


tests/Application/src/Entity/BookTranslationInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace App\Entity;
+
+use Sylius\Resource\Model\ResourceInterface;
+
+interface BookTranslationInterface extends ResourceInterface
```
GSadee, line 18:
```suggestion
interface BookTranslationInterface extends ResourceInterface, TranslationInterface
```

GSadee (approved):

GSadee:
Thank you, Loïc! :tada:

tests/Application/src/Entity/BookTranslationInterface.php
loic425: Yes, I've noticed that too. The AbstractTranslation already implements that TranslationInterface, so normally it will not change anything but I could change that to reflect better a real use case.

PR #942, opened 2024-09-23
Doctrine requirement is now optional

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,6 @@
         "babdev/pagerfanta-bundle": "^3.7 || ^4.0",
         "doctrine/annotations": "^2.0",
         "doctrine/collections": "^1.8 || ^2.0",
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -89,6 +89,7 @@
         "willdurand/hateoas-bundle": "^2.0"
     },
     "conflict": {
+        "doctrine/doctrine-bundle": "<2.0 || ^3.0",
```
loic425, line 92: cause it has been moved on dev requirement, we have to ensure the right version is installed.

GSadee (approved):

PR #986, opened 2025-02-20
Init resource class list factory with attributes

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Based on #985

### DIFF
--- a/src/Component/src/Metadata/Resource/Factory/AttributesResourceClassListFactory.php
+++ b/src/Component/src/Metadata/Resource/Factory/AttributesResourceClassListFactory.php
@@ -18,7 +18,7 @@ use Sylius\Resource\Metadata\Resource\ResourceClassList;
 use Sylius\Resource\Reflection\ClassReflection;
 
 /**
- * Creates a resource name collection from {@see AsResource} attributes.
+ * Creates a resource class list from {@see AsResource} attributes.
  *
  * @experimental
  */
--- a/src/Component/src/Metadata/Resource/Factory/AttributesResourceClassListFactory.php
+++ b/src/Component/src/Metadata/Resource/Factory/AttributesResourceClassListFactory.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Resource\Metadata\Resource\Factory;
+
+use Sylius\Resource\Metadata\AsResource;
+use Sylius\Resource\Metadata\Resource\ResourceClassList;
+use Sylius\Resource\Reflection\ClassReflection;
+
+/**
+ * Creates a resource name collection from {@see AsResource} attributes.
+ *
+ * @experimental
+ */
+final class AttributesResourceClassListFactory implements ResourceClassListFactoryInterface
+{
+    /**
+     * @param array{paths: string[]} $mapping
+     */
+    public function __construct(
+        private readonly array $mapping,
+        private readonly ?ResourceClassListFactoryInterface $decorated = null,
+    ) {
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function create(): ResourceClassList
+    {
+        /** @var class-string[] $classes */
+        $classes = [];
+
+        if ($this->decorated) {
+            foreach ($this->decorated->create() as $resourceClass) {
+                $classes[$resourceClass] = true;
+            }
+        }
+
+        $paths = $this->mapping['paths'] ?? [];
+
+        foreach (ClassReflection::getResourcesByPaths($paths) as $resourceClass) {
+            if ([] === ClassReflection::getClassAttributes($resourceClass, AsResource::class)) {
+                continue;
+            }
+
+            $classes[$resourceClass] = true;
+        }
+
+        return new ResourceClassList(array_keys($classes));
+    }
+}
--- a/src/Component/src/Metadata/Resource/Factory/ResourceClassListFactoryInterface.php
+++ b/src/Component/src/Metadata/Resource/Factory/ResourceClassListFactoryInterface.php
@@ -23,7 +23,7 @@ use Sylius\Resource\Metadata\Resource\ResourceClassList;
 interface ResourceClassListFactoryInterface
 {
     /**
-     * Creates the resource name collection.
+     * Creates the resource class list.
      */
     public function create(): ResourceClassList;
 }
--- a/src/Component/src/Metadata/Resource/Factory/ResourceClassListFactoryInterface.php
+++ b/src/Component/src/Metadata/Resource/Factory/ResourceClassListFactoryInterface.php
@@ -1,0 +1,29 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Resource\Metadata\Resource\Factory;
+
+use Sylius\Resource\Metadata\Resource\ResourceClassList;
+
+/**
+ * Creates a resource class list value object.
+ *
+ * @experimental
+ */
+interface ResourceClassListFactoryInterface
+{
+    /**
+     * Creates the resource name collection.
+     */
+    public function create(): ResourceClassList;
+}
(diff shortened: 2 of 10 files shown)


src/Component/src/Metadata/Resource/Factory/AttributesResourceNameCollectionFactory.php
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Resource\Metadata\Resource\Factory;
```
loic425: I think it will be better like this
```suggestion
namespace Sylius\Resource\Metadata\Resource\NameCollection\Factory;
```

WDYT @diimpp?

loic425: But in this, case we should move The ResourceNameCollection too, and moved all the service related to MetadataCollection factories in another namespace too.
I have an issue, cause I forgot some "Experimental" tags and then we need class and interface aliases for that. So maybe we could keep this coherent.

diimpp: Discussed in slack; Since NameCollection is just a VO, then extensive namespacing is not needed and current one looks good enough.

src/Bundle/Resources/config/services/metadata.xml
```diff
@@ -12,6 +12,10 @@
 -->
 
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <imports>
```
diimpp, line 15: Since `import` is introduced, all other "common" services from this file needs to be moved to sub-files as well.

src/Bundle/Resources/config/services/metadata/resource_name.xml
```diff
@@ -0,0 +1,26 @@
...
+ file that was distributed with this source code.
+
+-->
+
+<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+        <service id="sylius_resource.metadata.resource.name_collection.factory" alias="sylius_resource.metadata.resource.name_collection.factory.attributes" />
+        <service id="Sylius\Resource\Metadata\Resource\Factory\ResourceNameCollectionFactoryInterface" alias="sylius_resource.metadata.resource.name_collection.factory" />
+
+        <service id="sylius_resource.metadata.resource.name_collection.factory.attributes"
```
diimpp: Naming should match existing IDs.

Example
```yaml
sylius.resource_metadata_collection.factory.attributes
```

```suggestion
        <service id="sylius.resource_name_collection.factory.attributes"
```

loic425: So you prefer we change all the service ids in 2.0 than start using the new naming for the new service ids?

loic425:
Interesting discussions there
https://github.com/api-platform/core/pull/6943

But we are in a specific branch and also everything are tagged as experimental here, so we can adjust later.

src/Component/src/Metadata/Resource/Factory/AttributesResourceClassListFactory.php
```diff
@@ -0,0 +1,62 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Resource\Metadata\Resource\Factory;
+
+use Sylius\Resource\Metadata\AsResource;
+use Sylius\Resource\Metadata\Resource\ResourceClassList;
+use Sylius\Resource\Reflection\ClassReflection;
+
+/**
+ * Creates a resource name collection from {@see AsResource} attributes.
```
diimpp:
```suggestion
 * Creates a resource class list from {@see AsResource} attributes.
```

src/Component/src/Metadata/Resource/Factory/ResourceClassListFactoryInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Creates a resource class list value object.
+ *
+ * @experimental
+ */
+interface ResourceClassListFactoryInterface
+{
+    /**
+     * Creates the resource name collection.
```
diimpp:
```suggestion
     * Creates the resource class list.
```

src/Component/src/Metadata/Resource/ResourceClassList.php
```diff
@@ -14,11 +14,11 @@
...
- * A collection of resource names.
+ * A list of resource class names.
  *
  * @experimental
  */
-final class ResourceNameCollection implements \IteratorAggregate, \Countable
+final class ResourceClassList implements \IteratorAggregate, \Countable
 {
     /**
      * @param string[] $names
```
diimpp:
```patch
---     public function __construct(private readonly array $names = [])
+++     public function __construct(private readonly array $classNames = [])
+++     public function __construct(private readonly array $classes = [])
```

src/Component/tests/Metadata/Resource/Factory/AttributesResourceClassListFactoryTest.php
```diff
@@ -0,0 +1,34 @@
...
+    public function testCreateAResourceClassListForResourcesWithAsResourceAttribute(): void
+    {
+        $attributesResourceNameCollectionFactory = new AttributesResourceClassListFactory(
+            mapping: ['paths' => [dirname(__DIR__, 3) . '/Dummy']],
+        );
+
+        $collection = $attributesResourceNameCollectionFactory->create();
+
+        $this->assertContains(DummyResource::class, $collection->getIterator());
+        $this->assertNotContains(PullRequest::class, $collection->getIterator());
```
diimpp:
```suggestion
        $attributesResourceClassListFactory = new AttributesResourceClassListFactory(
            mapping: ['paths' => [dirname(__DIR__, 3) . '/Dummy']],
        );

        $list = $attributesResourceClassListFactory->create();

        $this->assertContains(DummyResource::class, $list->getIterator());
        $this->assertNotContains(PullRequest::class, $list->getIterator());
```

src/Component/tests/Metadata/Resource/ResourceClassListTest.php
```diff
@@ -14,27 +14,27 @@
...
 
-final class ResourceNameCollectionTest extends TestCase
+final class ResourceClassListTest extends TestCase
 {
     public function testItIsAnIteratorAggregate(): void
     {
-        $collection = new ResourceNameCollection();
+        $collection = new ResourceClassList();
 
         $this->assertInstanceOf(\IteratorAggregate::class, $collection);
```
diimpp:
```suggestion
        $list = new ResourceClassList();

        $this->assertInstanceOf(\IteratorAggregate::class, $list);
```

src/Component/tests/Metadata/Resource/ResourceClassListTest.php
```diff
@@ -14,27 +14,27 @@
...
     }
 
     public function testItIsCountable(): void
     {
-        $collection = new ResourceNameCollection();
+        $collection = new ResourceClassList();
 
         $this->assertInstanceOf(\Countable::class, $collection);
```
diimpp:
```suggestion
        $list = new ResourceClassList();

        $this->assertInstanceOf(\Countable::class, $list);
```

src/Component/tests/Metadata/Resource/ResourceClassListTest.php
```diff
@@ -14,27 +14,27 @@
...
     }
 
-    public function testItIsACollectionOfResourceNames(): void
+    public function testItIsAListOfResourceClassNames(): void
     {
-        $collection = new ResourceNameCollection(['first_resource', 'second_resource']);
+        $collection = new ResourceClassList(['first_resource', 'second_resource']);
```
diimpp:
```suggestion
        $list = new ResourceClassList(['first_resource', 'second_resource']);
```

src/Component/src/Metadata/Resource/ResourceClassList.php
diimpp: `src/Component/src/Metadata/Resource/Factory/AttributesResourceClassListFactory.php` uses `$classes` internally, so I guess it should be preferred.

src/Bundle/Resources/config/services/metadata/resource_name.xml
diimpp: @loic425 yes, but it's more about consistency -- two files in same directory shouldn't have different ID naming scheme.

So I guess `sylius_resource.***` can be left for 2.0.

diimpp (approved):

PR #998, opened 2025-03-12
Use plural name from new resource metadata for route paths

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

For now, we have two very similar metadata classes for the resource.
- [Metadata](https://github.com/Sylius/SyliusResourceBundle/blob/1.13/src/Component/src/Metadata/Metadata.php) 
- [ResourceMetadata](https://github.com/Sylius/SyliusResourceBundle/blob/1.13/src/Component/src/Metadata/ResourceMetadata.php)

To calculate the route path, we use the plural name from the Metadata (the legacy one), but we can define the plural name on the ResourceMetadata. 

Before this implementation, the plural name was calculated by the inflector on the MetadataClass.

With this current PR, by default, it will continue to use the plural name from the inflector (from the Medadata class), but it will use the custom one if defined. It was intended to be the case with the previous implementation, but there was an issue and the operation route factory was using the automatic plural name only.

The other plan behind is also to prepare the legacy Metadata class at some point.
We will be able to move the automatic plural name on the getResourceDefaults method.

### DIFF
--- a/src/Component/src/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
+++ b/src/Component/src/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
@@ -134,6 +134,12 @@ final class AttributesResourceMetadataCollectionFactory implements ResourceMetad
             $resource = $resource->withName($resourceConfiguration->getName());
         }
 
+        if (null === $resource->getPluralName()) {
+            $resourcePluralName = $resourceConfiguration->getPluralName();
+
+            $resource = $resource->withPluralName($resourcePluralName);
+        }
+
         return $resource;
     }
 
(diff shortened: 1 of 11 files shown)


src/Component/src/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
```diff
@@ -143,20 +149,6 @@ private function getOperationWithDefaults(ResourceMetadata $resource, Operation
 
         $operation = $operation->withResource($resource);
 
-        if (null === $resource->getName()) {
```
loic425, line 146: It's already handled on getResourceWithDefaults method. Here, it does not work corrrectly (see plural name).

src/Component/src/Symfony/Routing/Factory/OperationRouteFactory.php
```diff
@@ -30,9 +30,9 @@ public function __construct(
     ) {
     }
 
-    public function create(MetadataInterface $metadata, ResourceMetadata $resource, HttpOperation $operation): Route
+    public function create(ResourceMetadata $resource, HttpOperation $operation): Route
```
loic425, line 33: This class and interface are marked as experimental.

src/Component/src/Symfony/Routing/Factory/OperationRouteFactory.php
```diff
@@ -48,14 +48,17 @@ public function create(MetadataInterface $metadata, ResourceMetadata $resource,
...
-        return $this->getDefaultRoutePathForOperation($metadata, $operation);
+        return $this->getDefaultRoutePathForOperation($resource, $operation);
     }
 
-    private function getDefaultRoutePathForOperation(MetadataInterface $metadata, HttpOperation $operation): string
+    private function getDefaultRoutePathForOperation(ResourceMetadata $resource, HttpOperation $operation): string
     {
-        $rootPath = sprintf('%s', Urlizer::urlize($metadata->getPluralName()));
+        $pluralName = $resource->getPluralName();
+        Assert::notNull($pluralName, 'Plural name of the resource should be defined');
```
loic425, line 59: The plural name is populated previously with an automatic plural name.

stlgaits (approved):

PR #984, opened 2025-02-20
Add Operation defaults trait

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

PR #999, opened 2025-03-13
Use plural name from new resource metadata

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

It replaces #998 
For now, we have two very similar metadata classes for the resource.
- [Metadata](https://github.com/Sylius/SyliusResourceBundle/blob/1.13/src/Component/src/Metadata/Metadata.php) 
- [ResourceMetadata](https://github.com/Sylius/SyliusResourceBundle/blob/1.13/src/Component/src/Metadata/ResourceMetadata.php)

To calculate the route path, we use the plural name from the Metadata (the legacy one), but we can define the plural name on the ResourceMetadata. 

Before this implementation, the plural name was calculated by the inflector on the MetadataClass.

With this current PR, by default, it will continue to use the plural name from the inflector (from the Medadata class), but it will use the custom one if defined. It was intended to be the case with the previous implementation, but there was an issue and the operation route factory was using the automatic plural name only.

The other plan behind is also to prepare the legacy Metadata class removal at some point.
We will be able to move the automatic plural name on the getResourceDefaults method.

### DIFF
--- a/src/Component/src/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
+++ b/src/Component/src/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
@@ -134,6 +134,12 @@ final class AttributesResourceMetadataCollectionFactory implements ResourceMetad
             $resource = $resource->withName($resourceConfiguration->getName());
         }
 
+        if (null === $resource->getPluralName()) {
+            $resourcePluralName = $resourceConfiguration->getPluralName();
+
+            $resource = $resource->withPluralName($resourcePluralName);
+        }
+
         return $resource;
     }
 
@@ -142,20 +148,6 @@ final class AttributesResourceMetadataCollectionFactory implements ResourceMetad
         $resourceConfiguration = $this->resourceRegistry->get($resource->getAlias() ?? '');
 
         $operation = $operation->withResource($resource);
-
-        if (null === $resource->getName()) {
-            $resourceName = $resourceConfiguration->getName();
-
-            $resource = $resource->withName($resourceName);
-            $operation = $operation->withResource($resource);
-        }
-
-        if (null === $resource->getPluralName()) {
-            $resourcePluralName = $resourceConfiguration->getPluralName();
-
-            $resource = $resource->withPluralName($resourcePluralName);
-            $operation = $operation->withResource($resource);
-        }
 
         if (null === $operation->getNormalizationContext()) {
             $operation = $operation->withNormalizationContext($resource->getNormalizationContext());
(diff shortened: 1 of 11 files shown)


src/Component/src/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
```diff
@@ -134,6 +134,12 @@ private function getResourceWithDefaults(string $resourceClass, ResourceMetadata
             $resource = $resource->withName($resourceConfiguration->getName());
         }
 
+        if (null === $resource->getPluralName()) {
+            $resourcePluralName = $resourceConfiguration->getPluralName();
+
+            $resource = $resource->withPluralName($resourcePluralName);
+        }
```
diimpp, line 141: Can be optimized a little.

```suggestion
        if (null === $resource->getPluralName()) {
            $resource = $resource->withPluralName(
                $resourceConfiguration->getPluralName()
            );
        }
```

src/Component/src/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
```diff
@@ -143,20 +149,6 @@ private function getOperationWithDefaults(ResourceMetadata $resource, Operation
 
         $operation = $operation->withResource($resource);
 
-        if (null === $resource->getName()) {
```
diimpp, line 146: A question about line of code above this code.

```php
        $resourceConfiguration = $this->resourceRegistry->get($resource->getAlias() ?? '');
```

We probably should assert it as `Assert::notNull($resource->getAlias());` as was in some other PR/code.

src/Component/src/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactory.php
```diff
@@ -143,20 +149,6 @@ private function getOperationWithDefaults(ResourceMetadata $resource, Operation
...
-            $resourceName = $resourceConfiguration->getName();
-
-            $resource = $resource->withName($resourceName);
-            $operation = $operation->withResource($resource);
-        }
-
-        if (null === $resource->getPluralName()) {
```
diimpp, line 153: I guess plural name is set somehow else, so this code is removed, but I don't fully understand how or where.

diimpp (approved):

loic425:
This PR will need to be refactored after #984 merge.

loic425:
This has been replaced by #1108

PR #1004, opened 2025-03-14
[Metadata] Add PHP file resource extractor

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #1029
| License         | MIT

Same feature as I did on https://github.com/api-platform/core/pull/7017

### DIFF
--- a/src/Component/src/Metadata/Extractor/AbstractResourceExtractor.php
+++ b/src/Component/src/Metadata/Extractor/AbstractResourceExtractor.php
@@ -20,6 +20,8 @@ use Symfony\Component\DependencyInjection\ContainerInterface as SymfonyContainer
 
 /**
  * Base file extractor.
+ *
+ * @experimental
  */
 abstract class AbstractResourceExtractor implements ResourceExtractorInterface
 {
--- a/src/Component/src/Metadata/Extractor/AbstractResourceExtractor.php
+++ b/src/Component/src/Metadata/Extractor/AbstractResourceExtractor.php
@@ -1,0 +1,123 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Resource\Metadata\Extractor;
+
+use Psr\Container\ContainerInterface;
+use Sylius\Resource\Exception\RuntimeException;
+use Sylius\Resource\Metadata\ResourceMetadata;
+use Symfony\Component\DependencyInjection\ContainerInterface as SymfonyContainerInterface;
+
+/**
+ * Base file extractor.
+ */
+abstract class AbstractResourceExtractor implements ResourceExtractorInterface
+{
+    /** @var ResourceMetadata[]|null */
+    protected ?array $resources = null;
+
+    private array $collectedParameters = [];
+
+    /** @param string[] $paths */
+    public function __construct(
+        protected array $paths,
+        private readonly ?ContainerInterface $container = null,
+    ) {
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getResources(): array
+    {
+        if (null !== $this->resources) {
+            return $this->resources;
+        }
+
+        $this->resources = [];
+        foreach ($this->paths as $path) {
+            $this->extractFromPath($path);
+        }
+
+        return $this->resources;
+    }
+
+    /**
+     * Extracts metadata from a given path.
+     */
+    abstract protected function extractFromPath(string $path): void;
+
+    /**
+     * Recursively replaces placeholders with the service container parameters.
+     *
+     * @see https://github.com/symfony/symfony/blob/6fec32c/src/Symfony/Bundle/FrameworkBundle/Routing/Router.php
+     *
+     * @param mixed $value The source which might contain "%placeholders%"
+     *
+     * @throws \RuntimeException When a container value is not a string or a numeric value
+     *
+     * @return mixed The source with the placeholders replaced by the container
+     *               parameters. Arrays are resolved recursively.
+     */
+    protected function resolve(mixed $value): mixed
+    {
+        if (null === $this->container) {
+            return $value;
+        }
+
+        if (\is_array($value)) {
+            foreach ($value as $key => $val) {
+                $value[$key] = $this->resolve($val);
+            }
+
+            return $value;
+        }
+
+        if (!\is_string($value)) {
+            return $value;
+        }
+
+        $escapedValue = preg_replace_callback('/%%|%([^%\s]++)%/', function ($match) use ($value) {
+            $parameter = $match[1] ?? null;
+
+            // skip %%
+            if (!isset($parameter)) {
+                return '%%';
+            }
+
+            if (preg_match('/^env\(\w+\)$/', $parameter)) {
+                throw new RuntimeException(\sprintf('Using "%%%s%%" is not allowed in routing configuration.', $parameter));
+            }
+
+            if (\array_key_exists($parameter, $this->collectedParameters)) {
+                return $this->collectedParameters[$parameter];
+            }
+
+            if ($this->container instanceof SymfonyContainerInterface) {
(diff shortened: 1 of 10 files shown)


src/Component/src/Metadata/Extractor/PhpFileResourceExtractor.php
```diff
@@ -0,0 +1,53 @@
...
+
+        $resourceReflection = new \ReflectionClass($resource);
+
+        foreach ($resourceReflection->getProperties() as $property) {
+            $property->setAccessible(true);
+            $resolvedValue = $this->resolve($property->getValue($resource));
+            $property->setValue($resource, $resolvedValue);
+        }
+
+        $this->resources = [$resource];
```
diimpp: Function `public function getResources(): array` is meant to extract multiple resources by multiple paths, but this line overrides `$resources`.

Is this planned behavior?

src/Component/src/Metadata/Extractor/PhpFileResourceExtractor.php
```diff
@@ -0,0 +1,53 @@
...
+namespace Sylius\Resource\Metadata\Extractor;
+
+use Sylius\Resource\Metadata\ResourceMetadata;
+
+final class PhpFileResourceExtractor extends AbstractResourceExtractor
+{
+    /**
+     * @inheritdoc
+     */
+    protected function extractPath(string $path): void
```
diimpp: It's a resource extractor, not path extractor, so I guess `extractFromPath` is more suited name.

```suggestion
    protected function extractFromPath(string $path): void
```

src/Component/src/Metadata/Extractor/PhpFileResourceExtractor.php
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Sylius\Resource\Metadata\Extractor;
+
+use Sylius\Resource\Metadata\ResourceMetadata;
+
+final class PhpFileResourceExtractor extends AbstractResourceExtractor
+{
+    /**
+     * @inheritdoc
+     */
```
diimpp: I might be mistaken, but I think inheritdoc is an outdated phpdoc and static analyzer will understand it anyway. If it's true, then all instances of this needs to be removed in the PR.

```suggestion
```

src/Component/src/Metadata/Extractor/AbstractResourceExtractor.php
```diff
@@ -0,0 +1,120 @@
...
+                $resolved = $this->container?->get($parameter);
+            }
+
+            if (\is_string($resolved) || is_numeric($resolved)) {
+                $this->collectedParameters[$parameter] = $resolved;
+
+                return (string) $resolved;
+            }
+
+            throw new \RuntimeException(\sprintf('The container parameter "%s", used in the resource configuration value "%s", must be a string or numeric, but it is of type %s.', $parameter, $value, \gettype($resolved)));
```
diimpp: I think it should use `Resource\Exception\RuntimeException`

src/Component/src/Metadata/Extractor/AbstractResourceExtractor.php
```diff
@@ -0,0 +1,120 @@
...
+        $escapedValue = preg_replace_callback('/%%|%([^%\s]++)%/', function ($match) use ($value) {
+            $parameter = $match[1] ?? null;
+
+            // skip %%
+            if (!isset($parameter)) {
+                return '%%';
+            }
+
+            if (preg_match('/^env\(\w+\)$/', $parameter)) {
+                throw new \RuntimeException(\sprintf('Using "%%%s%%" is not allowed in routing configuration.', $parameter));
```
diimpp: Same

src/Component/src/Metadata/Extractor/ResourceExtractorInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+use Sylius\Resource\Exception\InvalidArgumentException;
+
+/**
+ * Extracts an array of metadata from a file or a list of files.
+ */
+interface ResourceExtractorInterface
+{
+    /**
+     * Parses all metadata files and convert them in an array.
```
diimpp, line 26: Would be cool to add `ResourceMetadata` return type.

src/Component/src/Metadata/Extractor/ResourceExtractorInterface.php
```diff
@@ -0,0 +1,29 @@
...
+     *
+     * @throws InvalidArgumentException
```
diimpp: There is no code in this PR, that throws `InvalidArgumentException` -- is it actual?

src/Component/src/Metadata/Extractor/AbstractResourceExtractor.php
loic425: done, you are right.

src/Component/src/Metadata/Extractor/AbstractResourceExtractor.php
loic425: also done.

src/Component/src/Metadata/Extractor/ResourceExtractorInterface.php
loic425: Right, I agree, done.

src/Component/src/Metadata/Extractor/ResourceExtractorInterface.php
loic425: Yes, copy paste from APIP, we do not throw that exception.

src/Component/src/Metadata/Extractor/PhpFileResourceExtractor.php
loic425: removed

src/Component/src/Metadata/Extractor/PhpFileResourceExtractor.php
loic425: yes, indeed, that's a better name.

src/Component/src/Metadata/Extractor/PhpFileResourceExtractor.php
loic425: you are right, this is really strange. I need to check that ASAP and on APIP too.

loic425: AHAH, ok, I remember now, the path is a specific file  (not a directory). So we have only one resource metadata per file. That's not the case we some other resource extractor on APIP where we can declare serveral resources in the same file.

loic425: we configure a directory on the config, but the Symfony Finder will extract files.

loic425: Have a look here
https://github.com/api-platform/core/blob/main/src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php#L427

diimpp: Right now if you have multiple file paths, each extract will override previous extract, so even if you have 5 different paths, you only will have single resource as output.

This change will fix that.

```patch
        $this->resources = [];
        foreach ($this->paths as $path) {
---         $this->extractFromPath($path);
       }
        $this->resources = [];
        foreach ($this->paths as $path) {
+++        $this->resources[] = $this->extractFromPath($path);
        }

Comment (loic425) [reply src/Component/src/Metadata/Extractor/PhpFileResourceExtractor.php]: Fix, thx a lot.

File: src/Component/src/Metadata/Extractor/AbstractResourceExtractor.php [hunk 8]
```diff
@@ -0,0 +1,123 @@
...
+    /** @param string[] $paths */
+    public function __construct(
+        protected array $paths,
+        private readonly ?ContainerInterface $container = null,
+    ) {
+    }
+
+    /**
+     * @inheritdoc
+     */
```
diimpp, line 42:
```suggestion
```

diimpp (approved):

loic425:
Test with the Skeleton are not working here but works in next PR :( I don't understand why, it works on my local :1st_place_medal:

PR #1013, opened 2025-05-05
Add new SecurityProvider to allow managing access to resource(s)

Prometee:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Add a Security provider to handle security access to a  resource operation.

- [x] Add `SecurityProvider`
- [x] Add tests
- [ ] Add documentation


src/Bundle/Resources/config/services/state/provider.xml
```diff
@@ -18,30 +18,39 @@
...
             <argument type="service" id="sylius.form.factory" />
         </service>
 
-        <service id="sylius.state_provider.validate" class="Sylius\Resource\Symfony\Validator\State\ValidateProvider" decorates="sylius.state_provider.read" decoration-priority="100">
+        <service id="sylius.state_provider.validate" class="Sylius\Resource\Symfony\Validator\State\ValidateProvider" decorates="sylius.state_provider.read" decoration-priority="200">
             <argument type="service" id=".inner" />
             <argument type="service" id="validator" />
         </service>
+
+        <service id="sylius.state_provider.security" class="Sylius\Resource\State\Provider\SecurityProvider" decorates="sylius.state_provider.read" decoration-priority="100">
```
loic425: I'm wondering if this should be a negative value to be after userland ones.

src/Bundle/Resources/config/services/state/provider.xml
```diff
@@ -18,30 +18,39 @@
...
+            <argument type="service" id=".inner" />
+            <argument type="service" id="sylius.state_provider.security_attribute" />
+            <argument type="service" id="security.authorization_checker" />
```
loic425: hum... Is this in optional requirements? in this case, you shoud handle this with NullOnInvalid

src/Component/src/Metadata/Operation.php
```diff
@@ -346,4 +347,17 @@ public function withNotificationMessage(string $notificationMessage): self
 
         return $self;
     }
+
+    public function canSecurity(): ?bool
```
loic425:
```suggestion
    public function hasSecurity(): ?bool
```

src/Component/src/State/Provider/SecurityAttributeProviderInterface.php
```diff
@@ -0,0 +1,25 @@
...
+
+use Sylius\Resource\Context\Context;
+use Sylius\Resource\Metadata\Operation;
+
+/**
+ * @experimental
+ */
+interface SecurityAttributeProviderInterface
+{
+    public function getAttribute(Operation $operation, Context $context): array|string;
```
loic425: To be equals to the Symfony argument name
```suggestion
    public function getAttributes(Operation $operation, Context $context): array|string;
```

src/Component/src/State/Provider/SecurityProvider.php
```diff
@@ -0,0 +1,53 @@
...
+        private readonly SecurityAttributeProviderInterface $securityAttributeProvider,
+        private AuthorizationCheckerInterface $authorizationChecker,
+    ) {
+    }
+
+    public function provide(Operation $operation, Context $context): object|array|null
+    {
+        $data = $this->provider->provide($operation, $context);
+
+        if (false === ($operation->canSecurity() ?? false)) {
```
loic425:
```suggestion
        if (false === ($operation->hasSecurity() ?? false)) {
```

loic425:
@Prometee Your PR is really cool, but we should think about the whole security system first.
I like the expression language on APIP too... and sth close to #[IsGranted] new Symfony attributes.

src/Bundle/Resources/config/services/state/provider.xml
Prometee: I will rewrite this part to copy the behavior of APIP : https://github.com/api-platform/core/blob/main/src/State/Provider/SecurityParameterProvider.php#L53

src/Component/src/Metadata/Operation.php
Prometee: It will become `getSecurity` in the end

Prometee:
I agree, I'm going to use the APIP way, there is a few changes to do from what I read.

src/Bundle/Resources/config/services/state/provider.xml
loic425: yes, but I think they removed the service when symfony/security is not available. There are two solution, remove it or set null if the service is invalid an throw an error to "try to run composer require ..."

Prometee: @loic425 you mean this : https://github.com/api-platform/core/blob/main/src/Symfony/Bundle/DependencyInjection/ApiPlatformExtension.php#L838-L849 ?

loic425 (approved):

src/Component/src/State/Provider/SecurityProvider.php
```diff
@@ -0,0 +1,50 @@
...
+use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+
+/**
+ * @experimental
+ */
+final class SecurityProvider implements ProviderInterface
+{
+    public function __construct(
+        private ProviderInterface $provider,
+        private OperationAccessCheckerInterface $operationAccessChecker,
```
diimpp:
```suggestion
        private readonly ProviderInterface $provider,
        private readonly OperationAccessCheckerInterface $operationAccessChecker,
```

src/Component/src/Symfony/Security/OperationAccessChecker.php
```diff
@@ -0,0 +1,89 @@
...
+            throw new \LogicException('The "symfony/security" library must be installed to use the "security" attribute.');
+        }
+
+        if (null === $this->expressionLanguage) {
+            throw new \LogicException('The "symfony/expression-language" library must be installed to use the "security" attribute.');
+        }
+
+        if (null === $expression = $operation->getSecurity()) {
+            return true;
+        }
```
diimpp: While I personally enjoy this way of getting and checking value at the same time -- it's considered a kind of anti-pattern lately and should be avoided used in public libraries.

```suggestion
        $expression = $operation->getSecurity();
        if (null === $expression) {
            return true;
        }
```

image

src/Component/src/Symfony/Security/OperationAccessChecker.php
```diff
@@ -0,0 +1,89 @@
...
+
+        $variables = array_merge($extraVariables, [
+            'trust_resolver' => $this->authenticationTrustResolver,
+            'auth_checker' => $this->authorizationChecker, // needed for the is_granted expression function
+        ]);
+
+        if (null === $token = $this->tokenStorage->getToken()) {
+            $token = new NullToken();
+        }
```
diimpp:
```suggestion
        $token = $this->tokenStorage->getToken();
        if (null === $token) {
            $token = new NullToken();
        }
```

src/Component/src/Symfony/Security/OperationAccessChecker.php
```diff
@@ -0,0 +1,89 @@
...
+
+        $variables = array_merge($variables, $this->getVariables($token));
```
diimpp: Why array over here
```php

        $variables = array_merge($extraVariables, [
            'trust_resolver' => $this->authenticationTrustResolver,
            'auth_checker' => $this->authorizationChecker, // needed for the is_granted expression function
        ]);
```

but method over here

```php
$this->getVariables($token)
```

```suggestion
        $variables = array_merge($variables, [
            'token' => $token,
            'user' => $token->getUser(),
            'roles' => $this->getEffectiveRoles($token),
        ]);
```

diimpp: 1. I'm not following why `getEffectiveRoles` is split from SF example `getVariables()`. 
2. Simple `getRoles()` name-wise should be enough, adding `Effective` suggests implementation details on this method, which are un-important in this context.

src/Component/src/Symfony/Security/OperationAccessChecker.php
diimpp, line 1: I would suggest to refactor this function in this way (similar to SF example, that you've provided)
```php
        if (null === $token = $this->tokenStorage->getToken()) {
            $token = new NullToken();
        }

        $variables = array_merge($extraVariables, $this->getVariables($token));
```
and

```php
    /**
     * @see https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Core/Authorization/Voter/ExpressionVoter.php
     */
    private function getVariables(TokenInterface $token): array
    {
        $roleNames = $token->getRoleNames();

        if (null !== $this->roleHierarchy) {
            $roleNames = $this->roleHierarchy->getReachableRoleNames($roleNames);
        }

        return [
            'token' => $token,
            'user' => $token->getUser(),
            'roles' => $roleNames,
            'trust_resolver' => $this->authenticationTrustResolver,
            'auth_checker' => $this->authorizationChecker, // needed for the is_granted expression function
        ];
    }
```

src/Component/src/Symfony/Security/OperationAccessChecker.php
Prometee: @diimpp this file is a copy of https://github.com/api-platform/core/blob/main/src/Symfony/Security/ResourceAccessChecker.php

src/Component/src/Symfony/Security/OperationAccessChecker.php
```diff
@@ -0,0 +1,87 @@
...
+
+        $expression = $operation->getSecurity();
+        if (null === $expression) {
+            return true;
+        }
+
+        $variables = array_merge($extraVariables, [
+            'trust_resolver' => $this->authenticationTrustResolver,
+            'auth_checker' => $this->authorizationChecker, // needed for the is_granted expression function
+        ]);
```
diimpp:
```suggestion
```

src/Component/src/Symfony/Security/OperationAccessChecker.php
```diff
@@ -0,0 +1,87 @@
...
+
+        $token = $this->tokenStorage->getToken();
+        if (null === $token) {
+            $token = new NullToken();
+        }
+
+        $variables = array_merge($variables, $this->getVariables($token));
```
diimpp:
```suggestion
        $variables = array_merge($extraVariables, $this->getVariables($token));
```

src/Component/src/Symfony/Security/OperationAccessChecker.php
```diff
@@ -0,0 +1,83 @@
...
+        if (null === $this->expressionLanguage) {
+            throw new \LogicException('The "symfony/expression-language" library must be installed to use the "security" attribute.');
+        }
+
+        $expression = $operation->getSecurity();
+        if (null === $expression) {
+            return true;
+        }
+
+
```
diimpp:
```suggestion
```

diimpp (approved):

NoResponseMate (approved):

PR #1028, opened 2025-06-04
[Metadata] Register metadata configuration

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no/yes
| Deprecations?   | no/yes 
| Related tickets | 
| License         | MIT

It replaces #1005

The goal of this PR is to achieve part of what have been done on https://github.com/api-platform/core/pull/7017


src/Bundle/Resources/config/services/metadata/php.xml
diimpp, line 1: Just an idea, personally I found service files following actual dir structure more cleaner.

```patch
--- src/Bundle/Resources/config/services/metadata/php.xml
+++ src/Bundle/Resources/config/services/metadata/extractor.xml
```

src/Bundle/DependencyInjection/Configuration.php
```diff
@@ -39,6 +39,9 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->arrayNode('mapping')
                     ->addDefaultsIfNotSet()
                     ->children()
+                        ->arrayNode('imports')
```
diimpp, line 42: Do you have example how it will be used? We now have both `imports` and `paths`, where `imports` are paths for php only configs?

src/Bundle/Resources/config/services/metadata/resource_class_list.xml
```diff
@@ -22,5 +22,15 @@
             <argument>%sylius.resource.mapping%</argument>
         </service>
         <service id="Sylius\Resource\Metadata\Resource\Factory\AttributesResourceClassListFactory" alias="sylius.metadata.resource_class_list.factory.attributes" />
+
+<!--        <service id="sylius.metadata.resource_class_list.factory.php_file"-->
```
diimpp: Necessary?

loic425: Oops should be removed on this PR, thx

src/Bundle/Resources/config/services/metadata/php.xml
loic425: Yes, I've followed APIP filename, but extractor will be better. I agree

src/Bundle/DependencyInjection/Configuration.php
loic425: Yes look at the APIP PR for the whole explanation. Basically we need files that are not in the autoloader.

src/Bundle/Resources/config/services/metadata/resource_class_list.xml
loic425: removed

src/Bundle/Resources/config/services/metadata/php.xml
loic425: renamed, thx a lot.

src/Bundle/DependencyInjection/SyliusResourceExtension.php
```diff
@@ -320,4 +324,43 @@ private function loadResources(array $loadedResources, ContainerBuilder $contain
...
+    private function registerMetadataConfiguration(ContainerBuilder $container, array $config): void
+    {
+        $resources = $this->getResourcesToWatch($container, $config);
+
+        $container->getDefinition('sylius.metadata.resource_extractor.php_file')->replaceArgument(0, $resources);
+    }
+
+    private function getResourcesToWatch(ContainerBuilder $container, array $config): array
+    {
+        $resources = [];
```
diimpp: It's doesn't work with resources, but rather resource php configuration files.

```suggestion
    private function getResourceFilesToWatch(ContainerBuilder $container, array $config): array
    {
        $resourceFiles = [];
```

diimpp (approved):

src/Bundle/DependencyInjection/SyliusResourceExtension.php
loic425: done, you are right.

diimpp (approved):

Issue #1029, opened 2025-06-04
External PHP file configuration for resource operations


loic425:
Very similar to https://github.com/api-platform/core/pull/7017
but in Sylius resource context (for HTML routing)

```php
<?php
// config/sylius/resources/speaker.php
declare(strict_types=1);

use Sylius\Resource\Metadata\Create;
use Sylius\Resource\Metadata\ResourceMetadata;
use Sylius\Resource\Metadata\Update;
use Sylius\Resource\Metadata\Operations;
use App\Entity\Speaker;

return new ResourceMetadata()
    ->withClass(Speaker::class)
    ->withOperations(new Operations([
        new Create(),
        new Update(),
    ]))
;
```
- [x] https://github.com/Sylius/SyliusResourceBundle/pull/1004
- [x] https://github.com/Sylius/SyliusResourceBundle/pull/1028
- [x] https://github.com/Sylius/SyliusResourceBundle/pull/1030
- [x] https://github.com/Sylius/SyliusResourceBundle/pull/1031

PR #1030, opened 2025-06-06
[Metadata] Add Php file resource class list factory

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | part of #1029
| License         | MIT

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -183,6 +183,7 @@ jobs:
             -
                 name: Create-project with skeleton
                 run: |
+                    set -x
                     composer create-project --ansi "symfony/skeleton:${{ matrix.skeleton }}" skeleton_app
                     cd skeleton_app
                     composer config extra.symfony.allow-contrib true
--- a/src/Component/src/Metadata/Resource/Factory/PhpFileResourceClassListFactory.php
+++ b/src/Component/src/Metadata/Resource/Factory/PhpFileResourceClassListFactory.php
@@ -1,0 +1,54 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Resource\Metadata\Resource\Factory;
+
+use Sylius\Resource\Metadata\Extractor\ResourceExtractorInterface;
+use Sylius\Resource\Metadata\Resource\ResourceClassList;
+
+/**
+ * Creates a resource class list from PHP configuration files.
+ *
+ * @experimental
+ */
+final class PhpFileResourceClassListFactory implements ResourceClassListFactoryInterface
+{
+    public function __construct(
+        private readonly ResourceExtractorInterface $phpFileResourceMetadataExtractor,
+        private readonly ?ResourceClassListFactoryInterface $decorated = null,
+    ) {
+    }
+
+    public function create(): ResourceClassList
+    {
+        $classes = [];
+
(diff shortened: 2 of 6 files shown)


src/Component/src/Metadata/Resource/Factory/PhpFileResourceClassListFactory.php
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Creates a resource class list from PHP configuration files.
+ *
+ * @experimental
+ */
+final class PhpFileResourceClassListFactory implements ResourceClassListFactoryInterface
+{
+    public function __construct(
+        private readonly ResourceExtractorInterface $metadataExtractor,
```
loic425: maybe `$resourceExtractor` or `$resourceMetadataExtractor`

src/Component/src/Metadata/Resource/Factory/PhpFileResourceClassListFactory.php
```diff
@@ -0,0 +1,57 @@
...
+        private readonly ?ResourceClassListFactoryInterface $decorated = null,
+    ) {
+    }
+
+    /**
+     * @inheritdoc
+     */
```
diimpp:
```suggestion
```

src/Component/src/Metadata/Resource/Factory/PhpFileResourceClassListFactory.php
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Creates a resource class list from PHP configuration files.
+ *
+ * @experimental
+ */
+final class PhpFileResourceClassListFactory implements ResourceClassListFactoryInterface
+{
+    public function __construct(
+        private readonly ResourceExtractorInterface $metadataExtractor,
+        private readonly ?ResourceClassListFactoryInterface $decorated = null,
```
diimpp, line 28: Is it possible for `.inner` to be empty within decoration process?

src/Component/src/Metadata/Resource/Factory/PhpFileResourceClassListFactory.php
diimpp: Since the whole point of the this class to call phpFileResourceMetataResourceExtractor, then would be cool to use in the name to understand it at glance without checking service definition.

```suggestion
        private readonly ResourceExtractorInterface $phpFileResourceMetadataExtractor,
```

src/Component/src/Metadata/Resource/Factory/PhpFileResourceClassListFactory.php
loic425: The way of having that nullable decorated is that we can switch the decorated and the decorator, even in userland if needed.

.github/workflows/build.yml
```diff
@@ -183,6 +183,7 @@ jobs:
             -
                 name: Create-project with skeleton
                 run: |
+                    set -x
```
loic425, line 186: I added back this debug mode and that works... strange behaviour here.
Maybe that's a luck here.

src/Component/src/Metadata/Resource/Factory/PhpFileResourceClassListFactory.php
loic425: removed

src/Component/src/Metadata/Resource/Factory/PhpFileResourceClassListFactory.php
loic425: Very verbose, but ok.

diimpp (approved):

PR #1031, opened 2025-06-06
Add PHP file resource metadata collection factory

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | partially #1029
| License         | MIT

### DIFF
--- a/src/Bundle/Resources/config/services/metadata/resource_metadata_collection.xml
+++ b/src/Bundle/Resources/config/services/metadata/resource_metadata_collection.xml
@@ -17,16 +17,28 @@
             <tag name="cache.pool" />
         </service>
 
+        <service id="sylius.resource_metadata_collection.factory" alias="sylius.resource_metadata_collection.factory.attributes" />
+        <service id="Sylius\Resource\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface" alias="sylius.resource_metadata_collection.factory.attributes" />
+
         <service id="sylius.resource_metadata_collection.factory.attributes" class="Sylius\Resource\Metadata\Resource\Factory\AttributesResourceMetadataCollectionFactory">
             <argument type="service" id="sylius.resource_registry" />
             <argument type="service" id="sylius.routing.factory.operation_route_name_factory" />
         </service>
-        <service id="Sylius\Resource\Metadata\Resource\Factory\ResourceMetadataCollectionFactoryInterface" alias="sylius.resource_metadata_collection.factory.attributes" />
-        <service id="sylius.resource_metadata_collection.factory" alias="sylius.resource_metadata_collection.factory.attributes" />
+
+        <service id="sylius.resource_metadata_collection.factory.php_file"
+                 class="Sylius\Resource\Metadata\Resource\Factory\PhpFileResourceMetadataCollectionFactory"
+                 decorates="sylius.resource_metadata_collection.factory"
+                 decoration-priority="400"
+        >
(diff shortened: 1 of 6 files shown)


src/Bundle/Resources/config/services/metadata/resource_metadata_collection.xml
```diff
@@ -17,16 +17,29 @@
             <tag name="cache.pool" />
         </service>
 
+        <service id="sylius.resource_metadata_collection.factory" alias="sylius.resource_metadata_collection.factory.attributes" />
```
diimpp, line 20: Is there reason to move existing alias over here? Usually alias created after service, which is aliased?

src/Bundle/Resources/config/services/metadata/resource_metadata_collection.xml
```diff
@@ -17,16 +17,29 @@
...
+                 class="Sylius\Resource\Metadata\Resource\Factory\PhpFileResourceMetadataCollectionFactory"
+                 decorates="sylius.resource_metadata_collection.factory"
+                 decoration-priority="400"
+        >
+            <argument type="service" id="sylius.resource_registry" />
+            <argument type="service" id="sylius.routing.factory.operation_route_name_factory" />
+            <argument type="service" id="sylius.metadata.resource_extractor.php_file" />
+            <argument type="service" id=".inner" />
+        </service>
+        <service id="Sylius\Resource\Metadata\Resource\Factory\PhpFileResourceMetadataCollectionFactory" alias="sylius.resource_metadata_collection.factory.php_file" />
```
diimpp: Shouldn't ID be some kind of an interface? Do we need alias to class itself?

src/Component/src/Metadata/Resource/Factory/PhpFileResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,76 @@
...
+
+        foreach ($this->phpfileResourceMetadataExtractor->getResources() as $resource) {
+            if ($resourceClass !== $resource->getClass()) {
+                continue;
+            }
+
+            $resourceAlias = $resource->getAlias();
+
+            if (null !== $resourceAlias) {
+                $resourceConfiguration = $this->resourceRegistry->get($resource->getAlias() ?? '');
```
diimpp:
```suggestion
            $resourceAlias = $resource->getAlias();
            if (null !== $resourceAlias) {
                $resourceConfiguration = $this->resourceRegistry->get($resourceAlias);
```

src/Bundle/Resources/config/services/metadata/resource_metadata_collection.xml
loic425: Yes, that's the main alias for all the decorators, and that's the main service injected in our routing loader.
We also did that here.
https://github.com/Sylius/SyliusResourceBundle/blob/1.14/src/Bundle/Resources/config/services/metadata/resource_class_list.xml#L16

loic425: I've moved the interface alias here too...

src/Bundle/Resources/config/services/metadata/resource_metadata_collection.xml
loic425: Yes, you are right, I've just removed that alias.

diimpp (approved):

PR #1041, opened 2025-06-20
[ExpressionLanguage] Add providers on argument parsers

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

```php
<?php

use Sylius\Resource\Metadata\Create;
use Sylius\Resource\Metadata\Operations;
use Sylius\Resource\Metadata\ResourceMetadata;

return (new ResourceMetadata())
    // ...
    ->withOperations(operations: new Operations(operations: [
        new Create(
            // ...
            factoryMethod: 'createForProduct',
            factoryArguments: [
                "throw_not_found_on_null(sylius_repositories.get('sylius.repository.product').find(request.attributes.get('productId'))), 'Product not found."
            ],
        ),
    ]))
;

```

### DIFF
--- a/src/Component/src/Symfony/ExpressionLanguage/ArgumentParser.php
+++ b/src/Component/src/Symfony/ExpressionLanguage/ArgumentParser.php
@@ -13,7 +13,9 @@ declare(strict_types=1);
 
 namespace Sylius\Resource\Symfony\ExpressionLanguage;
 
+use Symfony\Component\ExpressionLanguage\ExpressionFunctionProviderInterface;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
+use Webmozart\Assert\Assert;
 
 /**
  * @experimental
@@ -21,9 +23,15 @@ use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 final class ArgumentParser implements ArgumentParserInterface
 {
     public function __construct(
-        private ExpressionLanguage $expressionLanguage,
-        private VariablesCollectionInterface $variablesCollection,
+        private readonly ExpressionLanguage $expressionLanguage,
+        private readonly VariablesCollectionInterface $variablesCollection,
+        readonly ?iterable $providers = null,
     ) {
+        foreach ($providers ?? [] as $provider) {
+            Assert::isInstanceOf($provider, ExpressionFunctionProviderInterface::class);
+
+            $this->expressionLanguage->registerProvider($provider);
+        }
     }
 
     public function parseExpression(string $expression, array $variables = []): mixed
--- a/src/Component/src/Symfony/ExpressionLanguage/Provider/ThrowNotFoundOnNullExpressionFunctionProvider.php
+++ b/src/Component/src/Symfony/ExpressionLanguage/Provider/ThrowNotFoundOnNullExpressionFunctionProvider.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Resource\Symfony\ExpressionLanguage\Provider;
+
+use Symfony\Component\ExpressionLanguage\ExpressionFunction;
+use Symfony\Component\ExpressionLanguage\ExpressionFunctionProviderInterface;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+final class ThrowNotFoundOnNullExpressionFunctionProvider implements ExpressionFunctionProviderInterface
+{
+    public function getFunctions(): array
+    {
+        return [
+            new ExpressionFunction(
+                'throw_not_found_on_null',
+                function (string $value, string ...$args): string {
+                    $message = $args[0] ?? '';
+
+                    return sprintf(
+                        '(null !== %1$s) ? %1$s : throw new \Symfony\Component\HttpKernel\Exception\NotFoundHttpException(%2$s)',
+                        $value,
+                        $message,
+                    );
+                },
+                function (array $arguments, mixed $value, ?string $message = null): mixed {
+                    if (null === $value) {
+                        throw new NotFoundHttpException($message ?? '');
+                    }
+
+                    return $value;
+                },
+            ),
+        ];
+    }
+}
--- a/src/Bundle/Resources/config/services/expression_language.xml
+++ b/src/Bundle/Resources/config/services/expression_language.xml
@@ -42,19 +42,26 @@
             <argument type="tagged_iterator" tag="sylius.routing_variables" />
         </service>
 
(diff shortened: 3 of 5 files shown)


src/Component/src/Symfony/ExpressionLanguage/ArgumentParser.php
```diff
@@ -13,17 +13,25 @@
...
-        private VariablesCollectionInterface $variablesCollection,
+        private readonly ExpressionLanguage $expressionLanguage,
+        private readonly VariablesCollectionInterface $variablesCollection,
+        private readonly ?iterable $providers = null,
     ) {
+        Assert::allIsInstanceOf($providers ?? [], ExpressionFunctionProviderInterface::class);
+
+        foreach ($this->providers ?? [] as $provider) {
+            $this->expressionLanguage->registerProvider($provider);
+        }
```
diimpp: 2. ⚠️ Potential Issue: Assert::allIsInstanceOf(...)

    This line:

```php
Assert::allIsInstanceOf($providers ?? [], ExpressionFunctionProviderInterface::class);
```

is checking that all items in $providers implement ExpressionFunctionProviderInterface, which is good practice.

However, $providers might not be a traversable array (could be any iterable), and some assertion libraries (like Webmozart\Assert) expect an array in allIsInstanceOf.

So this might throw an error if $providers is a Traversable object (e.g., Generator, ArrayIterator, etc.).

    ✅ Fix suggestion: Convert $providers to an array before asserting:

```php
        Assert::allIsInstanceOf(is_iterable($providers) ? iterator_to_array($providers) : [], ExpressionFunctionProviderInterface::class);
```
        or if your assertion library supports iterable directly, you’re fine.

3. ⚠️ Potential Redundancy: $this->providers ?? []

    Since $this->providers is typed as ?iterable, foreach ($this->providers ?? []) is safe, but somewhat redundant.

    You could use:
```php
    foreach ($providers ?? [] as $provider) {
        $this->expressionLanguage->registerProvider($provider);
    }
```
    since $providers is the constructor parameter, and you're not mutating it.

4. ✅ General Good Practices

    Using dependency injection and early validation is good.

    Avoiding side effects outside the constructor (e.g., delayed registration) would be ideal, but it's acceptable here if this is a service setup step.

Summary of Recommendations

    ✔️ Convert $providers to array if your assertion library doesn't support generic iterables.

    ✔️ Consider using $providers directly in the loop if it's not mutated and clearer than using $this->providers.

    ❗ Be careful with assumptions on what iterable means — iterator_to_array() can help you ensure it’s safe to pass to functions expecting arrays.

src/Component/src/Symfony/ExpressionLanguage/Provider/NotFoundOnNullExpressionFunctionProvider.php
```diff
@@ -0,0 +1,48 @@
...
+
+final class NotFoundOnNullExpressionFunctionProvider implements ExpressionFunctionProviderInterface
+{
+    private const DEFAULT_MESSAGE = 'Requested page is invalid.';
+
+    public function getFunctions(): array
+    {
+        return [
+            new ExpressionFunction(
+                'not_found_on_null',
```
diimpp: I'm not too familiar with expression language, but
- Does it need `sylius_` prefix?
- Should it be `throw_not_found_on_null` to be more precise?

src/Component/src/Symfony/ExpressionLanguage/Provider/NotFoundOnNullExpressionFunctionProvider.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Sylius\Resource\Symfony\ExpressionLanguage\Provider;
+
+use Symfony\Component\ExpressionLanguage\ExpressionFunction;
+use Symfony\Component\ExpressionLanguage\ExpressionFunctionProviderInterface;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+final class NotFoundOnNullExpressionFunctionProvider implements ExpressionFunctionProviderInterface
+{
+    private const DEFAULT_MESSAGE = 'Requested page is invalid.';
```
diimpp: Does it need anything beyond `Not found.`?
But at any case, `invalid` shouldn't be used with not found code.

```suggestion
    private const DEFAULT_MESSAGE = 'Not found.';
```

src/Component/src/Symfony/ExpressionLanguage/ArgumentParser.php
loic425: fixed

src/Component/src/Symfony/ExpressionLanguage/Provider/NotFoundOnNullExpressionFunctionProvider.php
loic425: there are no built-in functions with expression language, so I don't think we need to prefix them.

loic425: I use your more precise name.

src/Component/src/Symfony/ExpressionLanguage/Provider/NotFoundOnNullExpressionFunctionProvider.php
loic425: yep, that was the message on the previous service, but this one is better, thx.

src/Component/src/Symfony/ExpressionLanguage/Provider/ThrowNotFoundOnNullExpressionFunctionProvider.php
```diff
@@ -0,0 +1,48 @@
...
+
+                    return sprintf(
+                        '(null !== %1$s) ? %1$s : throw new \Symfony\Component\HttpKernel\Exception\NotFoundHttpException(%2$s)',
+                        $value,
+                        $message,
+                    );
+                },
+                function (array $arguments, mixed $value, ?string $message = null): mixed {
+                    if (null === $value) {
+                        throw new NotFoundHttpException($message ?? self::DEFAULT_MESSAGE);
```
diimpp: Thinking about it, do we need default message at all? Why not just leave it empty?

diimpp (approved):

src/Component/src/Symfony/ExpressionLanguage/Provider/ThrowNotFoundOnNullExpressionFunctionProvider.php
loic425: Hum, good idea!

loic425: I'll change that tomorrow morning.

loic425: done

diimpp (approved):

PR #1050, opened 2025-07-08
Add route condition on new routing system

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

The system works for the bc-layer usage.
We just need to have two route names

```php
new Index(routeName: '_sylius_admin_product_index', grid: 'sylius_admin_product'),
```

**bc-layer disabled**
image

**bc-layer enabled**
image


src/Component/tests/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactoryTest.php
```diff
@@ -581,6 +582,51 @@ public function testItCreatesResourceMetadataWithRoutePrefix(): void
...
+        $this->assertInstanceOf(ResourceMetadataCollection::class, $metadataCollection);
+
+        $resource = $metadataCollection->getIterator()->current();
+        $this->assertInstanceOf(ResourceMetadata::class, $resource);
+        $this->assertSame('app.dummy', $resource->getAlias());
+
+        $operations = $resource->getOperations();
+        $this->assertInstanceOf(Operations::class, $operations);
+        $this->assertCount(4, $operations);
+        $this->assertTrue($operations->has('app_dummy_create'));
```
diimpp: Is it necessary to test resource route condition (in addition to operation route condition) as well or it's covered by other tests?

diimpp (approved):

src/Component/tests/Metadata/Resource/Factory/AttributesResourceMetadataCollectionFactoryTest.php
loic425: Yes, you're right. I'll remove those ones.

diimpp: @loic425 Hmm, I've actually meant, that since we have `operation` route condition test here, then wouldn't we need `resource` route condition test as well. :D 

But if those are not necessary, then it's all cool. :)

loic425: done, thx.

NoResponseMate (approved):

PR #1016, opened 2025-05-18
Fix ClassReflection resource class resolving edge case

Pierstoval:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | Related to #1017 , but is only a hotfix
| License         | MIT

This is only a hotfix but the overall resolution system must be re-thought.

PR #1052, opened 2025-07-25
fix:  class reflection was wrongly parsing docblocks

### LINKED ISSUE #1011: Docblock containing the word "class" causes wrong class loading in Sylius entity parsing
### Sylius version(s) affected

2.0.6

### Description

When a PHP class contains a docblock header where the word class appears before the actual class keyword (in the comment), Sylius mistakenly tries to load a non-existent class.
It seems that the parser incorrectly interprets the first occurrence of the word class found in the file, even if it's inside a comment block, leading to unexpected behavior.

This issue occurs during entity metadata reading or any process relying on class scanning, and results in an attempt to autoload a class named after the word immediately following class in the comment block.

### How to reproduce

Create a simple entity, e.g., MyClass.php:

```
<?php

namespace App\Entity;

/**
 * This docblock for this class explains what to do.
 */
class MyClass
{
}
```

Observe that Syliust tries to load a class named explains (taken from the comment text), causing an error like:

`Class "explains" does not exist`

### Possible Solution

You could fix this issue by improving the class detection in
ResourceBundle/src/Component/src/Reflection/ClassReflection.php (around line 49).

Currently, the code uses:
```
if (!preg_match('/class +([^{ ]+)/', $fileContent, $matches)) {
    // no class found
    continue;
}
```
However, this regex may match the word class inside a comment block (/** ... */), causing incorrect behavior.

👉 I suggest improving the parsing logic by checking that the match is not located within a comment block.

may be this kind of code :
```
if (preg_match('/class +([^{ ]+)/', $fileContent, $matches)) {
    // Add extra validation to ensure this is not inside a comment
    if (preg_match('/^\s*[\*\/]/m', $matches[0])) {
        // It's inside a comment, ignore
        continue;
    }

    // Otherwise proceed
}
```

### Additional Context

_No response_

stlgaits:
Attempt to fix https://github.com/Sylius/SyliusResourceBundle/issues/1011, following up on work already done here https://github.com/Sylius/SyliusResourceBundle/pull/1016

This prevents the issue of doc block comments containing the word class from mistakenly being parsed.
As mentioned in #1016 , this is only a quick fix, perhaps we could resort to https://github.com/nikic/PHP-Parser in the future instead? 

| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | [fixes #1011](https://github.com/Sylius/SyliusResourceBundle/issues/1011), [following #1016](https://github.com/Sylius/SyliusResourceBundle/pull/1016)
| License         | MIT

### DIFF
--- a/src/Component/src/Reflection/ClassReflection.php
+++ b/src/Component/src/Reflection/ClassReflection.php
@@ -46,7 +46,7 @@ final class ClassReflection
 
             $namespace = $matches[1] ?? null;
 
-            if (!preg_match('/class\s+(\w+)/', $fileContent, $matches)) {
+            if (!preg_match('/^(?:\s*(?:final|abstract|readonly)\s+)*class\s+(\w+)/mi', $fileContent, $matches)) {
                 // no class found
                 continue;
             }
--- a/src/Component/legacy/tests/Dummy/DummyClassWithDocBlock.php
+++ b/src/Component/legacy/tests/Dummy/DummyClassWithDocBlock.php
@@ -1,0 +1,21 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Resource\Tests\Dummy;
+
+/**
+ * This docblock for this class explains what to do.
+ */
+final class DummyClassWithDocBlock
+{
+}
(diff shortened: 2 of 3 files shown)


src/Component/src/Reflection/ClassReflection.php
```diff
@@ -46,7 +46,7 @@ public static function getResourcesByPath(string $path): iterable
 
             $namespace = $matches[1] ?? null;
 
-            if (!preg_match('/class\s+(\w+)/', $fileContent, $matches)) {
+            if (!preg_match('/^(?:\s*(?:final|abstract|readonly)\s+)*class\s+(\w+)/mi', $fileContent, $matches)) {
```
loic425, line 49: nope, we could start with class.

loic425:
Fixed in #1117

PR #1053, opened 2025-07-26
Use formtype if configured

dannyvw:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no 
| License         | MIT

If formType is configured on resource, use that instead of DefaultResourceType

### DIFF
--- a/src/Bundle/DependencyInjection/SyliusResourceExtension.php
+++ b/src/Bundle/DependencyInjection/SyliusResourceExtension.php
@@ -150,7 +150,7 @@ final class SyliusResourceExtension extends Extension implements PrependExtensio
                         'model' => $className,
                         'controller' => ResourceController::class,
                         'factory' => Factory::class,
-                        'form' => DefaultResourceType::class,
+                        'form' => $resourceMetadata->getFormType() ?? DefaultResourceType::class,
                     ],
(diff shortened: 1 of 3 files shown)


dannyvw:
@loic425 Is there something more needed for this PR?

loic425:
Hi @dannyvw, thx for the proposal.
For now, I don't know the exact future of the resource registration, so I prefer not to add things that will be hard to remove due to bc compatibility. Indeed, we do not need this for the new routing system (or tell me the use case).
The legacy routing usage will be removed from Sylius e-commerce soon (PRs are opened). 
I prefer to wait if we'll need this for the e-commerce part, but I don't think it will be useful like this. For a specific resource class, the metadata is a collection, not a single one.

dannyvw:
@loic425 This was for the use case like this https://github.com/Sylius/SyliusResourceBundle/pull/1053/changes#diff-2b9379a9762e7fb349fa0a805fec4a916a53c0de96b7a3a3f3a701f4e1364e32R18, if you configure the form type this will not be used and always fallback to a default form type.

loic425:
Sorry, but I do not agree, when configured on the attribute, it is used.
image

But of course, it will not be used by the resource controller if you still use it, but please don't you use it anymore.

dannyvw:
Yes this was for the resource controller, but I will closed it then.

PR #1055, opened 2025-07-29
feat(metadata) Customize Resource & operations

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

It's the same feature as [this APIP one](https://github.com/api-platform/core/pull/7213).
We need this feature for [this issue](https://github.com/Sylius/Sylius/issues/18212).

### DIFF
--- a/src/Bundle/DependencyInjection/SyliusResourceExtension.php
+++ b/src/Bundle/DependencyInjection/SyliusResourceExtension.php
@@ -23,9 +23,13 @@ use Sylius\Bundle\ResourceBundle\SyliusResourceBundle;
 use Sylius\Component\Resource\Factory\FactoryInterface as LegacyFactoryInterface;
 use Sylius\Resource\Factory\Factory;
 use Sylius\Resource\Factory\FactoryInterface;
+use Sylius\Resource\Metadata\AsOperationMutator;
 use Sylius\Resource\Metadata\AsResource;
+use Sylius\Resource\Metadata\AsResourceMutator;
 use Sylius\Resource\Metadata\Metadata;
+use Sylius\Resource\Metadata\OperationMutatorInterface;
 use Sylius\Resource\Metadata\ResourceMetadata;
+use Sylius\Resource\Metadata\ResourceMutatorInterface;
 use Sylius\Resource\Reflection\ClassReflection;
 use Sylius\Resource\State\ProcessorInterface;
 use Sylius\Resource\State\ProviderInterface;
@@ -34,6 +38,7 @@ use Sylius\Resource\Twig\Context\Factory\ContextFactoryInterface;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\Config\Loader\LoaderInterface;
 use Symfony\Component\Config\Resource\DirectoryResource;
+use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 use Symfony\Component\DependencyInjection\Exception\RuntimeException;
@@ -75,6 +80,32 @@ final class SyliusResourceExtension extends Extension implements PrependExtensio
 
         $this->loadPersistence($config['drivers'], $config['resources'], $loader, $container);
         $this->loadResources($config['resources'], $container);
+
+        $container->registerAttributeForAutoconfiguration(
+            AsResourceMutator::class,
+            static function (ChildDefinition $definition, AsResourceMutator $attribute, \ReflectionClass $reflector): void {
+                if (!is_a($reflector->name, ResourceMutatorInterface::class, true)) {
+                    throw new RuntimeException(\sprintf('Resource mutator "%s" should implement %s', $reflector->name, ResourceMutatorInterface::class));
+                }
+
+                $definition->addTag('sylius.resource_mutator', [
+                    'resourceClass' => $attribute->resourceClass,
+                ]);
+            },
+        );
+
+        $container->registerAttributeForAutoconfiguration(
+            AsOperationMutator::class,
+            static function (ChildDefinition $definition, AsOperationMutator $attribute, \ReflectionClass $reflector): void {
+                if (!is_a($reflector->name, OperationMutatorInterface::class, true)) {
+                    throw new RuntimeException(\sprintf('Operation mutator "%s" should implement %s', $reflector->name, OperationMutatorInterface::class));
+                }
+
+                $definition->addTag('sylius.operation_mutator', [
+                    'operationName' => $attribute->operationName,
+                ]);
+            },
+        );
 
         $container->registerForAutoconfiguration(ProviderInterface::class)
             ->addTag('sylius.state_provider')
--- a/src/Bundle/SyliusResourceBundle.php
+++ b/src/Bundle/SyliusResourceBundle.php
@@ -31,6 +31,7 @@ use Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler\WinzouStateMachine
 use Sylius\Bundle\ResourceBundle\DependencyInjection\PagerfantaExtension;
 use Sylius\Resource\Symfony\DependencyInjection\Compiler\DisableMetadataCachePass;
 use Sylius\Resource\Symfony\DependencyInjection\Compiler\FallbackToKernelDefaultLocalePass;
+use Sylius\Resource\Symfony\DependencyInjection\Compiler\MetadataMutatorPass;
 use Symfony\Component\DependencyInjection\Compiler\PassConfig;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
@@ -54,6 +55,7 @@ final class SyliusResourceBundle extends Bundle
         $container->addCompilerPass(new FallbackToKernelDefaultLocalePass());
         $container->addCompilerPass(new DoctrineContainerRepositoryFactoryPass());
         $container->addCompilerPass(new DoctrineTargetEntitiesResolverPass(new TargetEntitiesResolver()), PassConfig::TYPE_BEFORE_OPTIMIZATION, 1);
+        $container->addCompilerPass(new MetadataMutatorPass());
         $container->addCompilerPass(new RegisterFormBuilderPass());
         $container->addCompilerPass(new RegisterFqcnControllersPass());
         $container->addCompilerPass(new RegisterResourceRepositoryPass());
--- a/src/Component/src/Metadata/Mutator/OperationMutatorCollection.php
+++ b/src/Component/src/Metadata/Mutator/OperationMutatorCollection.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Resource\Metadata\Mutator;
+
+use Sylius\Resource\Metadata\OperationMutatorInterface;
+
+/**
+ * @internal
+ */
+final class OperationMutatorCollection implements OperationMutatorCollectionInterface
+{
+    /** @var array<string, list<OperationMutatorInterface>> */
+    private array $mutators = [];
+
+    /**
+     * Adds a mutator to the container for a given operation name.
+     */
+    public function add(string $operationName, OperationMutatorInterface $mutator): void
+    {
+        $this->mutators[$operationName][] = $mutator;
+    }
+
+    public function get(string $id): array
+    {
+        return $this->mutators[$id] ?? [];
+    }
+
+    public function has(string $id): bool
+    {
+        return isset($this->mutators[$id]);
+    }
+}
--- a/src/Component/src/Metadata/Mutator/OperationMutatorCollectionInterface.php
+++ b/src/Component/src/Metadata/Mutator/OperationMutatorCollectionInterface.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Resource\Metadata\Mutator;
+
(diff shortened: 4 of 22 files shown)


src/Component/src/Metadata/Mutator/OperationResourceMutatorCollection.php
```diff
@@ -0,0 +1,42 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Resource\Metadata\Mutator;
+
+use Sylius\Resource\Metadata\OperationMutatorInterface;
+
+/**
+ * @internal
+ */
+final class OperationResourceMutatorCollection implements OperationMutatorCollectionInterface
```
diimpp: It's Operation Mutator, but Operation Resource Mutator Collection -- how name was derived? 

If resource part is necessary, should it be

```patch
--- OperationResourceMutatorCollection
+++ ResourceOperationMutatorCollection
```
like so?

src/Component/src/Metadata/Mutator/OperationMutatorCollection.php
```diff
@@ -0,0 +1,42 @@
...
+
+    public function get(string $id): array
+    {
+        return $this->mutators[$id] ?? [];
+    }
+
+    public function has(string $id): bool
+    {
+        return isset($this->mutators[$id]);
+    }
```
diimpp, line 42:
```suggestion
    public function get(string $operationName): array
    {
        return $this->mutators[$operationName] ?? [];
    }

    public function has(string $operationName): bool
    {
        return isset($this->mutators[$operationName]);
    }
```

src/Component/src/Metadata/Mutator/OperationMutatorCollection.php
```diff
@@ -0,0 +1,42 @@
...
+final class OperationResourceMutatorCollection implements OperationMutatorCollectionInterface
+{
+    private array $mutators = [];
+
+    /**
+     * Adds a mutator to the container for a given operation name.
+     */
+    public function add(string $operationName, OperationMutatorInterface $mutator): void
+    {
+        $this->mutators[$operationName][] = $mutator;
```
diimpp, line 31: Just thinking out loud -- so it allows multiple mutators per operation name, but how does it ensure priority when mutators conflicting with each others.

src/Bundle/SyliusResourceBundle.php
```diff
@@ -54,6 +55,7 @@ public function build(ContainerBuilder $container): void
         $container->addCompilerPass(new FallbackToKernelDefaultLocalePass());
         $container->addCompilerPass(new DoctrineContainerRepositoryFactoryPass());
         $container->addCompilerPass(new DoctrineTargetEntitiesResolverPass(new TargetEntitiesResolver()), PassConfig::TYPE_BEFORE_OPTIMIZATION, 1);
+        $container->addCompilerPass(new MutatorPass());
```
diimpp: `Mutator` is not specific enough name when used outside of `Sylius\Resource\Metadata` dir.

For example,
```suggestion
        $container->addCompilerPass(new MetadataMutatorPass());
```

src/Component/src/Metadata/Mutator/OperationMutatorCollectionInterface.php
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Collection of Operation mutators to mutate Operation metadata.
+ */
+interface OperationMutatorCollectionInterface extends ContainerInterface
+{
+    /**
+     * @return list<OperationMutatorInterface>
+     */
+    public function get(string $id): mixed;
```
diimpp: Shouldn't it specify `add` and `has` as well?

src/Component/src/Metadata/Mutator/ResourceMutatorCollectionInterface.php
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Collection of Resource mutators to mutate ResourceMetadata.
+ */
+interface ResourceMutatorCollectionInterface extends ContainerInterface
+{
+    /**
+     * @return list<ResourceMutatorInterface>
+     */
+    public function get(string $id): array;
```
diimpp, line 29: Shouldn't it specify `add` and `has` as well?

src/Component/src/Metadata/Mutator/ResourceMutatorCollection.php
```diff
@@ -0,0 +1,39 @@
...
+
+    public function get(string $id): array
+    {
+        return $this->mutators[$id] ?? [];
+    }
+
+    public function has(string $id): bool
+    {
+        return isset($this->mutators[$id]);
+    }
```
diimpp, line 39:
```suggestion
    public function get(string $resourceClass): array
    {
        return $this->mutators[$resourceClass] ?? [];
    }

    public function has(string $resourceClass): bool
    {
        return isset($this->mutators[$resourceClass]);
    }
```

src/Component/src/Metadata/Mutator/ResourceMutatorCollection.php
```diff
@@ -0,0 +1,39 @@
...
+final class ResourceResourceMutatorCollection implements ResourceMutatorCollectionInterface
+{
+    private array $mutators;
+
+    public function add(string $resourceClass, ResourceMutatorInterface $mutator): void
+    {
+        $this->mutators[$resourceClass][] = $mutator;
+    }
+
+    public function get(string $id): array
```
diimpp, line 31: When resource mutator is not found, I would probably expect `null` value, not empty array, but since it's internal and probably used in some `foreach`, then it's okay as is.

src/Component/src/Metadata/Resource/Factory/MutatorResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private function mutateOperations(Operations $operations): Operations
+    {
+        $newOperations = new Operations();
+
+        /**
+         * @var Operation $operation
+         * @var string $key
+         */
```
diimpp: Is this part necessary? Doesn't `Operations` have any internal type hints to handle this?

src/Component/src/Metadata/Resource/Factory/MutatorResourceMetadataCollectionFactory.php
```diff
@@ -0,0 +1,80 @@
...
+        if ($this->decorated) {
+            $resourceMetadataCollection = $this->decorated->create($resourceClass);
+        }
+
+        $newMetadataCollection = new ResourceMetadataCollection();
+
+        /** @var ResourceMetadata $resource */
+        foreach ($resourceMetadataCollection as $resource) {
+            $resource = $this->mutateResource($resource, $resourceClass);
+            $operations = $this->mutateOperations($resource->getOperations() ?? new Operations());
```
diimpp, line 44: While it looks cool in the source code -- it will be calling `mutateOperations` on null values, when it's not necessary.

Personally I would prefer to add checks and don't call `mutateOperations` and `withOperation` when there are no operations.

Though, is it even possible to have resource without operations?

src/Component/src/Metadata/Mutator/ResourceResourceMutatorCollection.php
```diff
@@ -0,0 +1,39 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Resource\Metadata\Mutator;
+
+use Sylius\Resource\Metadata\ResourceMutatorInterface;
+
+/**
+ * @internal
+ */
+final class ResourceResourceMutatorCollection implements ResourceMutatorCollectionInterface
```
diimpp: Naming scheme doesn't make sense to me, but I'm a bit of out the loop on metadata

src/Component/src/Metadata/Mutator/ResourceMutatorCollection.php
```diff
@@ -0,0 +1,39 @@
...
+namespace Sylius\Resource\Metadata\Mutator;
+
+use Sylius\Resource\Metadata\ResourceMutatorInterface;
+
+/**
+ * @internal
+ */
+final class ResourceResourceMutatorCollection implements ResourceMutatorCollectionInterface
+{
+    private array $mutators;
```
Prometee: **🔴 Critical: Uninitialized property**

The `$mutators` property is declared but never initialized. This will cause issues when trying to access it before any mutators are added.

```suggestion
private array $mutators = [];
```

Note: `OperationResourceMutatorCollection` correctly initializes this property on line 23 of that file.

src/Bundle/DependencyInjection/SyliusResourceExtension.php
```diff
@@ -76,6 +81,40 @@ public function load(array $configs, ContainerBuilder $container): void
         $this->loadPersistence($config['drivers'], $config['resources'], $loader, $container);
         $this->loadResources($config['resources'], $container);
 
+        $container->registerAttributeForAutoconfiguration(
+            AsResourceMutator::class,
+            static function (ChildDefinition $definition, AsResourceMutator $attribute, \Reflector $reflector): void {
+                if (!$reflector instanceof \ReflectionClass) {
+                    return;
+                }
```
Prometee: What about throwing an exception here to explain why it is required?

Prometee (changes requested):

src/Component/src/Metadata/Mutator/OperationResourceMutatorCollection.php
loic425: I think this was an issue during name refactoring at some point. We should rename them on APIP too.

src/Component/src/Metadata/Mutator/OperationMutatorCollection.php
loic425: we cannot do that, cause we break the name defined in the Container PSR interface.
image

Moreover, this collection is internal and so it's not intended to be used out of the package.

src/Component/src/Metadata/Mutator/OperationMutatorCollection.php
loic425: Yes, we have already thought about that (with soyuka), and it can be managed by setting a priority on the php attribute.

src/Component/src/Metadata/Mutator/OperationMutatorCollectionInterface.php
loic425: The PSR container interface already defines the `has` method.

We do not use the `add` method on abstraction, we only use the `get` method.
image

src/Component/src/Metadata/Mutator/ResourceMutatorCollection.php
loic425: yep, it's used on a `foreach`, it's marked as internal.
Maybe should we throw an exception later.

src/Bundle/DependencyInjection/SyliusResourceExtension.php
loic425: Yep, I've updated the code related to [Nico's comment](https://github.com/api-platform/core/pull/7213#discussion_r2381225425)

src/Component/src/Metadata/Mutator/OperationResourceMutatorCollection.php
loic425: I've fixed these 2 class names.

src/Bundle/SyliusResourceBundle.php
loic425: fixed

src/Component/src/Metadata/Mutator/ResourceMutatorCollectionInterface.php
loic425: see previous comment

src/Component/src/Metadata/Mutator/ResourceResourceMutatorCollection.php
loic425: fixed

src/Component/src/Metadata/Mutator/ResourceMutatorCollection.php
loic425: fixed

src/Bundle/DependencyInjection/SyliusResourceExtension.php
loic425: Oops, I did another one, but not this one, fixed

src/Component/src/Metadata/Resource/Factory/MutatorResourceMetadataCollectionFactory.php
loic425: done

Prometee (approved):

src/Component/src/Metadata/Resource/Factory/MutatorResourceMetadataCollectionFactory.php
loic425: In theory, yes we can have a resource without any operations.

src/Component/src/Metadata/Mutator/OperationMutatorCollectionInterface.php
```diff
@@ -0,0 +1,28 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Resource\Metadata\Mutator;
+
+use Psr\Container\ContainerInterface;
+use Sylius\Resource\Metadata\OperationMutatorInterface;
+
+/**
+ * Collection of Operation mutators to mutate Operation metadata.
```
loic425, line 20: I need to add an "experimental" tag.

src/Component/src/Metadata/Mutator/OperationMutatorCollectionInterface.php
loic425: I've marked these interfaces as experimental, so we'll be able to improve it if necessary.

src/Component/src/Metadata/Mutator/OperationMutatorCollectionInterface.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @experimental
+ */
+interface OperationMutatorCollectionInterface extends ContainerInterface
+{
+    /**
+     * @return list<OperationMutatorInterface>
+     */
+    public function get(string $id): mixed;
```
diimpp: Resource mutator collection return typed as `array`. Should it be the same?

```php
    /**
     * @return list<ResourceMutatorInterface>
     */
    public function get(string $id): array;
```

```suggestion
    public function get(string $id): array;
```

diimpp (approved):

diimpp (approved):

NoResponseMate (approved):

PR #1057, opened 2025-07-31
[Fix] Do not dispatch a post write event when data is already a response

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

When data returned by the write processor is already a response, we does not have to dispatch a post write event.

image


src/Component/src/Metadata/ResourceMetadata.php
```diff
@@ -28,9 +28,9 @@ public function __construct(
         private ?string $pluralName = null,
         private ?string $applicationName = null,
         private ?string $identifier = null,
-        protected ?array $normalizationContext = null,
-        protected ?array $denormalizationContext = null,
-        protected ?array $validationContext = null,
+        private ?array $normalizationContext = null,
```
loic425, line 31: strange... the ecs-fix did that change

loic425: but normally, nobody is using this object directly right now and this class is final.

NoResponseMate (approved):

PR #1066, opened 2025-09-24
Use Nikic php parser to find resources in files

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Cause we totally change the way to detect resource classes in files, I think it's safer to add this a new Sylius Resource release.

Note:
It seems to break memory usage.


composer.json
```diff
@@ -27,12 +27,14 @@
     "require": {
         "php": "^8.1",
         "babdev/pagerfanta-bundle": "^4.4",
+        "behat/transliterator": "^1.5",
```
loic425, line 30: waiting for upmerge

loic425:
Replaced by #1117

PR #1069, opened 2025-10-22
Fix CI

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -88,7 +88,7 @@
         "vimeo/psalm": "^5.26",
         "willdurand/hateoas-bundle": "^2.5",
         "winzou/state-machine-bundle": "^0.6.2",
-        "zenstruck/foundry": "^2.3"
+        "zenstruck/foundry": "^2.3 <2.7"
     },
     "conflict": {
         "doctrine/orm": "<2.18",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -88,7 +88,7 @@
         "vimeo/psalm": "^5.26",
         "willdurand/hateoas-bundle": "^2.5",
         "winzou/state-machine-bundle": "^0.6.2",
-        "zenstruck/foundry": "^2.3"
+        "zenstruck/foundry": "^2.3 <2.7"
```
loic425, line 91: quick fix here, we need to open an issue on Foundry side.
It's on dev requirement, so nothing important here.

GSadee: Why didn't you add the conflict? We usually use this approach with an appropriate description of why it was added in `CONFLICTS.md`.

GSadee (approved):

composer.json
loic425: Conflict is for main requirements (or optional ones), cause if I add a conflict, it will prevent users to use latest version of foundry.
This is an issue on test application only.

PR #1074, opened 2025-10-27
PHPSpec to PHPUnit DependencyInjection

Glancu:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT


loic425:
Please split into 2 commits:
- first one with only the changes on the same file
- second one with the files moved and renamed.

GSadee:
> Please split into 2 commits:
> 
> * first one with only the changes on the same file
> * second one with the files moved and renamed.

Hi @loic425! Just to better understand your reasoning, could you explain why this level of separation into two commits is important in your view? From my perspective, it feels a bit excessive in this particular case, especially since the PR is already quite granular and focused only on specific directory. I'm curious what concrete benefits you see in splitting it this way. 🤔

loic425:
When splitting into 2 commits we can review the diff in the first commit.
With @diimpp, we love this way, it's way faster to review.

Look at the current PRs inside the Grid package, and you'll see the diff.

On my OpenCode, I've split this into two tasks.

Rafikooo (approved):

PR #1089, opened 2025-11-10
Fix parameter %resource% for translation

### LINKED ISSUE #1062: Idea: Make it possible to translate only the key of the resource in flash message
### Description

Hello!

Currently in Sylius, when i change my backend to german and i for example update a product, it says:

"Product wurde erfolgreich aktualisiert".

As you can see "Product" is not correctly translated.

I dont know if there is a way to only translate the key of the resource (Product) in this case.

I only found a way to translate the hole resource flash message string with:

```yaml
sylius:
 resource: My string
```

But this affects every resource.

So far i dont found a way to translate only the resource key.

### Example

_No response_

michalkaczmarek-bitbag:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | https://github.com/Sylius/SyliusResourceBundle/issues/1062 
| License         | MIT

de_DE
Screenshot from 2025-11-27 22-22-07

fr_FR
Screenshot from 2025-11-27 22-23-33

pl_PL
Screenshot from 2025-11-27 22-24-22

### DIFF
--- a/src/Bundle/Controller/FlashHelper.php
+++ b/src/Bundle/Controller/FlashHelper.php
@@ -24,6 +24,8 @@ use Symfony\Contracts\Translation\TranslatorInterface;
 
 final class FlashHelper implements FlashHelperInterface
 {
+    private const UI_TRANSLATION_PREFIX = 'sylius.ui.';
+
     /** @var RequestStack|SessionInterface */
     private $requestStack;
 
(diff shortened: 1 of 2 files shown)


src/Bundle/Controller/FlashHelper.php
```diff
@@ -148,9 +148,22 @@ private function isTranslationDefined(string $message, string $locale, array $pa
     private function getParametersWithName(MetadataInterface $metadata, string $actionName): array
     {
         if (stripos($actionName, 'bulk') !== false) {
-            return ['%resources%' => ucfirst($metadata->getPluralName())];
+            $pluralName = $this->convertToSnakeCase($metadata->getPluralName());
+            $resourceKey = sprintf('sylius.ui.%s', $pluralName);
```
Rafikooo: Extract this prefix as `private const UI_TRANSLATION_PREFIX = 'sylius.ui.';`

Rafikooo (commented):
I would suggest extracting a dedicated private method, something like this:

```php
  private function getParametersWithName(MetadataInterface $metadata, string $actionName): array
  {
      if (stripos($actionName, 'bulk') !== false) {
          $resourceName = $metadata->getPluralName();
          $fallback = ucfirst($resourceName);

          return ['%resources%' => $this->translateResourceName($resourceName, $fallback)];
      }

      $resourceName = $metadata->getName();
      $fallback = ucfirst($metadata->getHumanizedName());

      return ['%resource%' => $this->translateResourceName($resourceName, $fallback)];
  }

  private function translateResourceName(string $resourceName, string $fallback): string
  {
      $snakeCaseName = $this->convertToSnakeCase($resourceName);
      $translationKey = sprintf('%s%s', self::UI_TRANSLATION_PREFIX, $snakeCaseName);
      $translated = $this->translator->trans($translationKey, [], 'messages');

      return $translated === $translationKey ? $fallback : $translated;
  }

``` 
And I’d suggest adding a photo to the PR description

Rafikooo (approved):

PR #1104, opened 2025-11-28
Fix static analyse, build docker image, and step related with removing GridBundle package  for Symfony 7.4

michalkaczmarek-bitbag:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -115,3 +115,17 @@ parameters:
         -
             message: '#If condition is always false#'
             path: src/Bundle/Event/ResourceControllerEvent.php
+        # Symfony <7.4 - TreeBuilder and ArrayNodeDefinition are not generic in Symfony < 7.4
+        # but are generic in Symfony 7.4+. We use generics for better type safety in newer versions.
+        - message: '#^PHPDoc tag .+ is not generic\.$#'
+          identifier: generics.notGeneric
+          count: 11
+          path: src/Bundle/DependencyInjection/Configuration.php
+        - message: '#^PHPDoc tag .+ is not generic\.$#'
+          identifier: generics.notGeneric
+          count: 5
+          path: src/Bundle/DependencyInjection/PagerfantaConfiguration.php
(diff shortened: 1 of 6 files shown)


phpstan.neon
```diff
@@ -115,3 +115,13 @@ parameters:
...
             message: '#If condition is always false#'
             path: src/Bundle/Event/ResourceControllerEvent.php
+
+        # Symfony Config Component generics compatibility
+        # Symfony 7.4+ introduced generic types for TreeBuilder and ArrayNodeDefinition
+        # Symfony 6.4 doesn't have these generic types yet
+        # We ignore the "missing generic type specification" errors from Symfony 7.4+
+        # and don't use generic type annotations to maintain compatibility with Symfony 6.4
+        # See: src/Bundle/DependencyInjection/Configuration.php
+        - '/Method .+::getConfigTreeBuilder\(\) return type with generic class Symfony\\Component\\Config\\Definition\\Builder\\TreeBuilder does not specify its types: T/'
```
Prometee: Hi @michalkaczmarek-bitbag !

We already did this kind of change into SyliusGridBundle like this:
https://github.com/Sylius/SyliusGridBundle/blob/1.15/phpstan.neon.dist#L21-L35
And we also introduce the right PHPDoc: 
https://github.com/Sylius/SyliusGridBundle/blob/1.15/src/Bundle/DependencyInjection/Configuration.php

Can we standardize this to keep the same way of doing thing from one repo to another?

Prometee (changes requested):

phpstan.neon
michalkaczmarek-bitbag: Hi @Prometee , thank you for your comment :) Yes, I changed according to your code

Prometee (approved):

Rafikooo (approved):

PR #1105, opened 2025-11-28
Bugfix/pin symfony 64 and 73 versions

michalkaczmarek-bitbag:
| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT


loic425:
Thx for the proposal, it's not needed anymore on 1.14 branch.
But yes, I agree with the idea. But for now, I think it's worth to introduce the prefer-lowest tests.

PR #1117, opened 2025-12-18
Add new reflection system for class names

### LINKED ISSUE #1017: Bug: ClassReflection resource resolution system is flawed
**Sylius version affected**: 1.13.0-ALPHA.3

The `\Sylius\Resource\Reflection\ClassReflection::getResourcesByPath()` method is using regexes, while it could find other ways, like reflection.

I just stumbled upon a strange error today regarding this:

```
In ClassReflection.php line 79:

  Class "App\Entity\Field\=" does not exist

```

The reflection is looping over several files that are actually `traits`, therefore it doesn't match and doesn't add them as resources, however, one of them has this code inside it:

```php
// ...
$class = ...
// ...
```

This triggers the current resolving process to actually consider this class as being a resource, and the matcher will resolve it to `=` (since it's what's after `class `. This is the cause of the error.

Can be easy-fixed by https://github.com/Sylius/SyliusResourceBundle/pull/1016 , though it's probably gonna break another app in the future.

Example that breaks after the hotfix of the aforementioned PR:

```php
// ...

/**
 * this class is used for ...
 */
class Id
{
// ...
```

Same here: this will break (I tested it) but the debug message will be easier to read (it will display the beginning of the comment, so it's supposedly fine).

IMO, autoloading, leveraging PSR-4 maybe, and/or class reflection should be used, as they will be more reliable.

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes (but introduce a deprecation is a new feature, according to Symfony rules)
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #1017
| License         | MIT

The system becomes from API-Platform source code. We know this system is already reliable.
https://github.com/api-platform/core/blob/main/src/Metadata/Util/ReflectionClassRecursiveIterator.php

### DIFF
--- a/src/Component/src/Reflection/ReflectionClassRecursiveIterator.php
+++ b/src/Component/src/Reflection/ReflectionClassRecursiveIterator.php
@@ -16,6 +16,10 @@ namespace Sylius\Resource\Reflection;
 /**
  * Gets reflection classes for php files in the given directories.
  *
+ * This class in inspired by this API Platform one:
+ *
+ * @see https://github.com/api-platform/core/blob/main/src/Metadata/Util/ReflectionClassRecursiveIterator.php
+ *
  * @internal
  */
 final class ReflectionClassRecursiveIterator
--- a/src/Component/src/Reflection/ReflectionClassRecursiveIterator.php
+++ b/src/Component/src/Reflection/ReflectionClassRecursiveIterator.php
@@ -1,0 +1,88 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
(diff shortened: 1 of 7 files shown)


src/Component/src/Reflection/ReflectionClassRecursiveIterator.php
```diff
@@ -0,0 +1,88 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Resource\Reflection;
+
+/**
+ * Gets reflection classes for php files in the given directories.
+ *
+ * @internal
+ */
+final class ReflectionClassRecursiveIterator
```
NoResponseMate, line 25: Would be nice to add a link to the OG class from APIP.

NoResponseMate (approved):

src/Component/src/Reflection/ReflectionClassRecursiveIterator.php
loic425: done, what do you think about the sentence?

PR #1130, opened 2026-01-05
Fix resource configuration for AsResource with only applicationName set

esserj:
it should not assume app, but only fall back on app if applicationName is not specified

| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

### DIFF
--- a/src/Bundle/DependencyInjection/SyliusResourceExtension.php
+++ b/src/Bundle/DependencyInjection/SyliusResourceExtension.php
@@ -216,6 +216,7 @@ final class SyliusResourceExtension extends Extension implements PrependExtensio
     private function getResourceAlias(ResourceMetadata $resource, string $className): string
     {
         $alias = $resource->getAlias();
+        $applicationName = $resource->getApplicationName() ?? 'app';
 
         if (null !== $alias) {
             return $alias;
@@ -229,7 +230,7 @@ final class SyliusResourceExtension extends Extension implements PrependExtensio
             $shortName = substr($shortName, 0, strlen($shortName) - strlen($suffix));
         }
 
-        return 'app.' . u($shortName)->snake()->toString();
+        return u($applicationName)->snake()->toString() . '.' . u($shortName)->snake()->toString();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


loic425 (approved):

loic425:
@NoResponseMate I'm wondering if we should introduce this to the 1.14 branch. Indeed it can result to a few issues, cause the service ids will be different for users who have configured a custom application name in their project.

esserj:
@loic425 @NoResponseMate Yes, and also any referencing to app.<resource> would no longer work that would have worked before.
I thought this can be seen as "misconfiguration" (setting it but then using `app` anyways) that is "easily" fixed by removing the `applicationName:` config from your `AsResource`, but it might indeed be better of in a minor release to not break any apps that need an urgent security release.

loic425:
Yes, I agree, so, could you rebase your PR with 1.14 branch please?

esserj:
done

loic425:
> done

Thx, that's perfect.

src/Bundle/DependencyInjection/SyliusResourceExtension.php
```diff
@@ -229,7 +230,7 @@ private function getResourceAlias(ResourceMetadata $resource, string $className)
             $shortName = substr($shortName, 0, strlen($shortName) - strlen($suffix));
         }
 
-        return 'app.' . u($shortName)->snake()->toString();
+        return u($applicationName)->snake()->toString() . '.' . u($shortName)->snake()->toString();
```
Prometee, line 233: @loic425 what about using sprintf to understand clearly what will be the resulted string?

```suggestion
        return sprintf('%s.%s',
                u($applicationName)->snake()->toString(),
                u($shortName)->snake()->toString(),
        );
```

Prometee (approved):

src/Bundle/DependencyInjection/SyliusResourceExtension.php
loic425: Yep, but in this case that's not a lot more readable, cause we have a simple concatenation here, not a complex one.

NoResponseMate (approved):

esserj:
Thanks for your time and effort @loic425 @Prometee @NoResponseMate

PR #1043, opened 2025-06-25
Add argument parser for resource and operation vars

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

The feature we need with the new routing system

**Previous implementation:**
```yaml
sylius_admin_catalog_promotion_product_variant_index:
    path: /catalog-promotions/{id}/variants
    defaults:
        _sylius:
            vars:
                catalogPromotion: expr:service('sylius.repository.catalog_promotion').find($id)
```

**New implementation:**
```php
new Index(vars: [
    'catalogPromotion' => "@=sylius_repositories.get('sylius.repository.catalog_promotion').find(request.attributes.get('id')",
]);
```

**Note**
`@=` is the synthax for expression that is used on Symfony dependency injection and Twig Hooks

Example on Sylius E-commerce admin panel
The product is used on the breadcrumb.

image

PR #1131, opened 2026-01-06
Deprecate expressions without prefix

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

Recently, we added an argument parser to use expression language on the "vars" for the resource routing using `@=` (https://github.com/Sylius/SyliusResourceBundle/pull/1043).
In the redirect and in the factory, we used an hack to detect if we need to parse the value or not.
Now it's time to use the same syntax with `@=`.

### DIFF
--- a/src/Component/src/State/Factory.php
+++ b/src/Component/src/State/Factory.php
@@ -74,6 +74,7 @@ final class Factory implements FactoryInterface
                 trigger_deprecation('sylius/resource-bundle', '1.14', 'You passed "%s" as a string value in your repository arguments. If this is a value that needs to be parsed using the expression language, please prefix your string with "@=". In your case, use "@=%s"."', $value, $value);
             }
 
+            // Not reachable as long as the BC layer above is there
             if (!str_starts_with($value, '@=')) {
                 $arguments[$key] = $value;
 
--- a/src/Component/src/State/Factory.php
+++ b/src/Component/src/State/Factory.php
@@ -69,6 +69,18 @@ final class Factory implements FactoryInterface
     private function parseArgumentValues(array $arguments): array
     {
         foreach ($arguments as $key => $value) {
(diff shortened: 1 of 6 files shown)


src/Component/src/State/Factory.php
```diff
@@ -69,6 +69,18 @@ public function create(Operation $operation, Context $context): ?object
...
+            if (!str_starts_with($value, '@=')) {
+                $value = '@=' . $value;
+                trigger_deprecation('sylius/resource-bundle', '1.14', 'You passed "%s" as a string value in your repository arguments. If this is a value that needs to be parsed using the expression language, please prefix your string with "@=". In your case, use "@=%s"."', $value, $value);
+            }
+
+            if (!str_starts_with($value, '@=')) {
+                $arguments[$key] = $value;
+
+                continue;
+            }
```
NoResponseMate, line 82: The second condition doesn't seem to be reachable

loic425: Yep, for now it's not reachable (with this bc layer enabled), I should add a comment, cause it's normal, it will not parse the expression if it's not prefixed with `@=` after bc-layer removal.

loic425: Maybe that'd be cool to build sth clever with a bc-layer configuration on the bundle for that.
`parse_prefixed_expressions_only: true`

loic425: @NoResponseMate For now, I've just added a comment on these non-reachable conditions.

NoResponseMate: I think just having the deprecation notice in place should suffice

NoResponseMate (approved):

PR #1132, opened 2026-01-13
fix translatable resource not getting its translation mapped with AsResource

### LINKED ISSUE #1018: Add support for translation entity association in AsResource attribute
Hello,

I noticed that it is currently not possible to associate a translation entity using the `AsResource` attribute.

In YAML, this kind of configuration is straightforward — we can define the translation class directly — but I couldn’t find an equivalent way to achieve this using PHP attributes.

esserj:
`AsResource` does not have `translations` configuration
and there seems to be no way to currently get it mapped properly when using attributes?
This ensures that when you `AsResource` tag your translation resource as `<alias>_translation`,
the mapping occurs, even without the resource metadata translation parameter.
which seemed to be an artificial limitation also not implemented in the mapTranslation

| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #1018
| License         | MIT

### DIFF
--- a/src/Bundle/EventListener/ORMTranslatableListener.php
+++ b/src/Bundle/EventListener/ORMTranslatableListener.php
@@ -88,6 +88,10 @@ final class ORMTranslatableListener implements EventSubscriber
      */
     private function mapTranslatable(ClassMetadata $metadata): void
     {
+        if ($metadata->hasAssociation('translations')) {
+            return;
+        }
+
         $className = $metadata->name;
 
         try {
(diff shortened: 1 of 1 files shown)


src/Bundle/EventListener/ORMTranslatableListener.php
```diff
@@ -96,24 +100,18 @@ private function mapTranslatable(ClassMetadata $metadata): void
             return;
         }
 
-        if (!$resourceMetadata->hasParameter('translation')) {
-            return;
-        }
```
esserj, line 101: seems to be an artificial limitation?

src/Bundle/EventListener/ORMTranslatableListener.php
```diff
@@ -96,24 +100,18 @@ private function mapTranslatable(ClassMetadata $metadata): void
...
-
         /** @var MetadataInterface $translationResourceMetadata */
         $translationResourceMetadata = $this->resourceMetadataRegistry->get($resourceMetadata->getAlias() . '_translation');
```
esserj, line 104: we could also try catch here, but it seems to make sense that when you mark your entity translatable, you'd want an error that it cannot find your translation resource

PR #1134, opened 2026-01-23
[New routing] Redirect to referer

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT


src/Component/src/Metadata/HttpOperation.php
```diff
@@ -52,6 +52,7 @@ public function __construct(
         string|\Stringable|null $security = null,
         ?string $securityMessage = null,
         string|callable|null $twigContextFactory = null,
+        protected string|null $redirectTo = null,
```
NoResponseMate:
```suggestion
        protected ?string $redirectTo = null,
```
to be consistent w/ the rest

src/Component/src/Symfony/Routing/RedirectHandler.php
```diff
@@ -31,6 +31,8 @@
  */
 final class RedirectHandler implements RedirectHandlerInterface
 {
+    private const REFERER = 'referer';
```
NoResponseMate: Maybe let's expose this through the interface since we're expecting an exact value

src/Component/src/Symfony/Routing/RedirectHandler.php
```diff
@@ -39,6 +41,23 @@ public function __construct(
...
+    public function redirect(
+        mixed $data,
+        HttpOperation $operation,
+        Request $request,
+    ): RedirectResponse {
+        if (self::REFERER === $operation->getRedirectTo()) {
+            /** @var string|null $referer */
+            $referer = $request->headers->get('referer');
+
+            if (null !== $referer and $this->isValidReferer($referer, $request)) {
```
NoResponseMate:
```suggestion
            if (null !== $referer && $this->isValidReferer($referer, $request)) {
```
💃

src/Component/src/Symfony/Routing/RedirectHandler.php
loic425: done, thanks a lot. 
I've applied the other suggestions, I agree with all of them.

src/Component/src/Symfony/Routing/RedirectHandlerInterface.php
```diff
@@ -22,6 +22,8 @@
  */
 interface RedirectHandlerInterface
 {
+    public const REFERER = 'referer';
```
loic425, line 25: I'm wondering if this is the right place, cause it's related to Symfony here.
But "referer" keywork is a generic one.

loic425: @NoResponseMate WDYT?

src/Component/src/Symfony/Routing/RedirectHandler.php
loic425: I added a question about this "referer" keyword constant.

src/Component/src/Symfony/Routing/RedirectHandlerInterface.php
NoResponseMate: We could extract it completely outside since the set value is based on an RFC, and there's really no generic place atm where it fits perfectly; Though I'm not sure it's really worth it

loic425: Yeah ok, you are right. So we can keep this constant here at least for now.

NoResponseMate (approved):

PR #1137, opened 2026-01-29
Fix redirect to referer

loic425:
| Q               | A
| --------------- | -----
| Bug fix?        | yes (but on 1.14 new feature)
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Component/src/Symfony/Routing/RedirectHandler.php
+++ b/src/Component/src/Symfony/Routing/RedirectHandler.php
@@ -39,11 +39,8 @@ final class RedirectHandler implements RedirectHandlerInterface
     ) {
     }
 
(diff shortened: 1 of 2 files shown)


src/Component/src/Symfony/Routing/RedirectHandler.php
```diff
@@ -39,11 +39,8 @@ public function __construct(
     ) {
     }
 
-    public function redirect(
```
loic425, line 42: This method is not used and not in the interface. Moreover, this class is marked as experimental.

NoResponseMate (approved):

PR #1154, opened 2026-02-27
Fix sourceEntity not updated when copying parent association mappings

mpysiak:
| Q               | A
|-----------------|-----
| Branch?         | 1.14
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

## Summary

When `ORMMappedSuperClassSubscriber` copies association mappings from a parent mapped-superclass to a child entity, the `sourceEntity` property is not updated to reflect the concrete child class. This causes Doctrine's `BasicEntityPersister` to generate incorrect SQL table aliases for eager-loaded inverse one-to-one associations.

**Root cause:** The subscriber directly assigns the parent's association mapping to the child metadata without updating `sourceEntity`, unlike Doctrine's own `ClassMetadataFactory::addInheritedRelations()` which sets `$mapping->sourceEntity = $subClass->name`.

**Symptom:** When the owning side of an inverse one-to-one association is eager-loaded, `BasicEntityPersister::getSelectColumnsSQL()` calls `getSQLTableAlias($association->sourceEntity, $assocAlias)` which produces a different alias than the one used for the JOIN table (because `sourceEntity` still points to the mapped-superclass instead of the concrete entity). This results in `Unknown column` SQL errors.

**Reproduction scenario:**
1. A mapped-superclass `BaseVendor` has a one-to-one owning association to `Profile`
2. A concrete entity `AppVendor extends BaseVendor` is registered via `resolve_target_entities`
3. `Profile` has an inverse one-to-one back to `VendorInterface` (resolved to `AppVendor`)
4. Loading `Profile` triggers eager-loading of the inverse side, which generates broken SQL because `sourceEntity` is `BaseVendor` instead of `AppVendor`

### DIFF
--- a/tests/Bundle/EventListener/ORMMappedSuperClassSubscriberTest.php
+++ b/tests/Bundle/EventListener/ORMMappedSuperClassSubscriberTest.php
@@ -18,48 +18,55 @@ use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Event\LoadClassMetadataEventArgs;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\DefaultNamingStrategy;
+use Doctrine\Persistence\Mapping\ClassMetadata as ClassMetadataInterface;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 use Doctrine\Persistence\Mapping\RuntimeReflectionService;
 use PHPUnit\Framework\TestCase;
 use Sylius\Bundle\ResourceBundle\EventListener\ORMMappedSuperClassSubscriber;
 use Sylius\Bundle\ResourceBundle\Tests\Fixtures\ChildEntity;
 use Sylius\Bundle\ResourceBundle\Tests\Fixtures\ParentEntity;
-use Sylius\Resource\Metadata\RegistryInterface;
+use Sylius\Resource\Metadata\Registry;
 
 final class ORMMappedSuperClassSubscriberTest extends TestCase
 {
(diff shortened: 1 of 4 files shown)


tests/Bundle/EventListener/ORMMappedSuperClassSubscriberTest.php
```diff
@@ -0,0 +1,77 @@
...
+use Sylius\Bundle\ResourceBundle\EventListener\ORMMappedSuperClassSubscriber;
+use Sylius\Bundle\ResourceBundle\Tests\Fixtures\ChildEntity;
+use Sylius\Bundle\ResourceBundle\Tests\Fixtures\ParentEntity;
+use Sylius\Resource\Metadata\RegistryInterface;
+
+final class ORMMappedSuperClassSubscriberTest extends TestCase
+{
+    public function testItUpdatesSourceEntityWhenCopyingParentAssociationMappings(): void
+    {
+        $registry = $this->createMock(RegistryInterface::class);
```
loic425: Could you try not using mocks please.

loic425: An example of mocks' removal
https://github.com/Sylius/SyliusResourceBundle/pull/1140/changes#diff-711921626c171bd19656c698b72785644c2ebeb0d7e5e6f88e9ab65fc06acbaaR38

mpysiak: Hi, I've fixed the tests 🖖

loic425 (approved):

NoResponseMate (approved):

PR #1158, opened 2026-03-03
Fix: preserve explicit class from #[AsResource] in getResourceWithDef…

### LINKED ISSUE #1157: Bug: explicit `class` parameter in `#[AsResource]` is always overwritten by the resource PHP class
## Description

  When using `#[AsResource(class: SomeEntity::class)]` to explicitly set the entity class on a Resource DTO, the `class` parameter is silently ignored. `ResourceMetadata::getClass()` always returns the Resource PHP class itself (e.g. `FruitResource`) instead of the explicitly configured entity class (e.g. `Fruit`).

  ## Steps to reproduce

  ```php
  #[AsResource(
      alias: 'app.fruit',
      class: Fruit::class,
      driver: false,
  )]
  final class FruitResource implements ResourceInterface {}

  $collection = $factory->create(FruitResource::class);
  $metadata->getClass(); // Expected: Fruit — Actual: FruitResource
```

## Root cause

```php
OperationDefaultsTrait::getResourceWithDefaults() unconditionally overwrites the class property with $resourceClass, ignoring the value from #[AsResource(class:)]:

 // line 34 — always overwrites
 $resource = $resource->withClass($resourceClass);
```

 This is inconsistent with every other property in the same method, all of which use null-checks before applying defaults (alias, applicationName, name).

 ## Proposed fix

 ```php
- $resource = $resource->withClass($resourceClass);
+ if (null === $resource->getClass()) {
+     $resource = $resource->withClass($resourceClass);
+ }
```

 ## Versions

  Reproduced on v1.14.0 and branch 1.15.

nfacciolo:
…aults()

| Q               | A
| --------------- | -----
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no-->
| Related tickets | fixes #1157
| License         | MIT

### DIFF
--- a/src/Component/src/Metadata/Resource/Factory/OperationDefaultsTrait.php
+++ b/src/Component/src/Metadata/Resource/Factory/OperationDefaultsTrait.php
@@ -31,7 +31,9 @@ trait OperationDefaultsTrait
         ResourceMetadata $resource,
         MetadataInterface $resourceConfiguration,
     ): ResourceMetadata {
-        $resource = $resource->withClass($resourceClass);
+        if (null === $resource->getClass()) {
+            $resource = $resource->withClass($resourceClass);
+        }
 
         if (null === $resource->getAlias()) {
             $resource = $resource->withAlias($resourceConfiguration->getAlias());


loic425:
@nfacciolo Thx for your proposal, please read the comment on the issue.

